PR #7, opened 2023-12-07
added theme switch for dark/light

LTDakin:
Changes
- ThemeSwitch component for toggling between dark and light mode

Another Note
- not final position of toggle, can be placed in other components or another place on the page as we add more

https://github.com/LCOGT/datalab-ui/assets/54085254/0d11e025-e600-4f7d-9abb-c01b6f631d89

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,5 +1,6 @@
 <script setup>
 import { ref, provide } from 'vue';
+import ThemeSwitch from './components/ThemeSwitch.vue';
 
 
 const selectedImages = ref([]);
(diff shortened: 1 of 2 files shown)


capetillo:
Lovely UI, but one quick comment, shouldn't the text switch to 'Light Mode' when dark mode is enabled? Just a lil thought

capetillo (commented):
Code LGTM! I would add an if statement in the themeswitch component's template where v-if toggleTheme is dark, then display text to be "light mode" and vice versa. But otherwise, code LGTM. My suggestion can also wait : )

LTDakin:
> Code LGTM! I would add an if statement in the themeswitch component's template where v-if toggleTheme is dark, then display text to be "light mode" and vice versa. But otherwise, code LGTM. My suggestion can also wait : )

Added some code to label the switch the reverse for now. So if its in light mode it will say light mode and if its in dark mode it will say dark mode. That makes sense, but also I don't know if it should say the opposite. Regardless it should be easy to change so just going to have it as reactive for now.

PR #9, opened 2023-12-18
Feature/image carousel

capetillo:
### Description:

**Background:** 
We want DataLab users to have the ability to switch views when viewing their Projects. The views are `Image Carousel` and `Image List`. When the user selects `Image Carousel` display, they can see their projects as image thumbnails, have their selected image displayed bigger, have the selected thumbnails have a yellow border and the ability to unselect them.
I also merged a branch with this one where I created Mock Data in JSON format. This Mock Data is currently being used as placeholder for real data. More Mock Data may be added. 

**Implementation:**

The `Image Carousel` component currently has a reactive `state` in it, but it will be moved to either a newly created store or the parent component. In the meantime, the `state` does two things: 1. Keeps track of `currentSlide` and 2. Stores `selectedImages` in an array. 

_currentSlide_: The first image (the one with an index of 0) will be displayed when the user first navigates to a project. If the user selects a different image, `currentSlide` updates to the index of the selected one so that one becomes the displayed image. This is done by two-way binding the `Carousel` element with the `currentSlide`. Lastly, if the user unselects a previously selected image, `currentImage` will become the last image stored in the `selectedImages` array.
_selectedImages_: This array gets updated when thumbnail images are selected or unselected. 

The `handleThumbnailClick` method  is invoked when a user clicks a thumbnail image and it does two things: 1. Toggles the status of thumbnail images to add and remove from the `selectedImages` array in state and 2. Updates `currentSlide` to display the most recently selected image. 

_handleThumbnailClick_: First, I use the `findIndex` method to see if the clicked image is already in the `selectedImages` array. If it is, then it gets removed from the array. If it isn't, it gets added to it. Then, it checks if `selectedImages` has length to update `currentSlide` so that the appropriate image is displayed bigger. 

The last method is `isSelected` and this is used to dynamically add a class when a thumbnail image is clicked so that they can have a yellow border when selected. 

### Visuals:

Some images are selected and have a yellow outline
Screenshot 2023-12-18 at 1 14 03 PM


LTDakin (approved):
lgtm

PR #10, opened 2023-12-19
Feature/archive list

LTDakin:
Implementation of a ImageList component for the project view page. This is an alternative to the Image Carousel. It takes a prop of mock data, the parent state of selectedImages, and emits the event when a new image is selected to update it's parents state. Also some light css and comment changes to maintain organization and appearance.

Screenshot 2023-12-18 at 4 09 35 PM


capetillo (approved):
Nice work, Lloyd!  As we've discussed, I'll create a vuex store and some of this will have to be rearranged, but nothing major (hopefully). Anyway, LGTM!

PR #11, opened 2023-12-20
Feature/datasession skeleton

jnation3406:


LTDakin (approved):
lgtm, but note to remove the api endpoint and token in the future

capetillo (approved):
For the record, we discussed the token's display and agreed that it's okay to leave it cause it's fake. Looks good to me

PR #13, opened 2023-12-22
Feature/import button

capetillo:
### DESCRIPTION:

**Background:**

We need an `import button` that POSTs `selectedImages` to the Mock API so that we can GET from said API in the `DataSession` view and display `selectedImages`

**Implementation:**

Since `selectedImages` is in the vuex store, we can access it globally now. Hooray! So, we access these and map through the array of `selectedImages` and pass `source` and `basename`. Then we set the `requestBody` to pass `name` (currently made up) and `input_data` which is the`selectedImages` that was mapped. Then the response is a standard fetch where the `method` is `POST`, the `headers` include the token (which is fake, no one will hack us) and the `Content-Type` is the same as Jon's had in his. And lastly, the `body` is the `requestBody`. Then there's standard error catching. And that's pretty much it. 

### VISUALS:
**Screenshot of data being passed:**
Screenshot 2023-12-22 at 2 38 55 PM


src/components/ImportButton.vue
```diff
@@ -1,22 +1,49 @@
...
+import { useStore } from 'vuex'
+
+const router = useRouter()
+const store = useStore()
+
+const importImages = async () => {
+  try {
+    const selectedImages = store.state.selectedImages
+    const inputData = selectedImages.map(image => ({
+      'source': image.image,
```
jnation3406, line 12: the source key being `image.image` seems a bit confusing, but maybe that is tangential to this PR. I would think a property on the image like `source` would make more sense? I know its all placeholder data for now though...

jnation3406 (approved):
Looks okay... Next steps I think are hooking the project import component into the main /datasessions page so it's not a separate route.

src/components/ImportButton.vue
capetillo: oh yeah that totally makes sense. The mock data has the images saved as images. I'll change that after break if that's okay : )

PR #15, opened 2024-01-04
created home page with nav

LTDakin:
added a HomeView component to have a landing page for the data lab. In the future we may decide to drop users straight into their last data session but for now this gives us an easy way instead of typing into the browser.


capetillo (approved):
Nicely done and so quick wow! LGTM

PR #17, opened 2024-01-04
Feature/project sidebar

capetillo:
### DESCRIPTION

**Background:**
I accidentally removed the project sidebar in my previous merge. I now not only reincorporated it again, but I refactored the code so that the sidebar is its own component. 

**Implementation:**
Refactored existing code for sidebar and I added it to the `ProjectBar` component. I also added the same logic to `handleProjectSelection` to both `ImageCarousel` and `ProjectList`. This logic could live in `ProjectBar` and be reused. But I want to have something for today's casual demo. I can fix this after merging too : )

### VISUALS 

**ProjectBar in Image Carousel view**
Screenshot 2024-01-04 at 11 34 19 AM

**ProjectBar in Project List view**

Screenshot 2024-01-04 at 11 34 54 AM


LTDakin:
Looks good, just would want ImageCarousel and ProjectBar to be in the projectview components folder

LTDakin (approved):
thanks looks good

PR #19, opened 2024-01-08
Feature/datasession-image-display

capetillo:
### DESCRIPTION

**Background:**
We need to display the `selectedImages` in the DataSession View that were POSTed to the api by clicking the `Add to Session` button in the Project View so that users can perform image operations on their images.

**Implementation:**
There's a new `ref` in the component `DataSession` and I named it `images` and initialized it to be an empty array. 
In `getImages()` I make a `GET` request to the mock api and assign the response (specifically, `data.results`) to the value of `images` to access these in the template and display them as images. I filter out any ones whose `source` includes the word "archive" because those won't display an image. 
To display the images, I loop through the `images` array in a `for...in` loop in the template. Then I require them from the `assets` local folder and the files are each of the `images`. The screenshots below will help convey my message.

**Important Note:**
Currently, when the user navigates to a different session (clicks on a separate tab), the first image to load is black. I'm not sure why this is, but I did add a conditional rendering to make sure that if there aren't any images to display, then don't display anything. 
Another important note is that all of the images from the api are being displayed for each data session rather than only the images associated with that session. Trying to find a workaround, but likely, the workaround will be that we need to implement functionality for which session we want to add selectedImages to.

### VISUALS
**Console: response data, screen: no first image**
Screenshot 2024-01-08 at 3 14 43 PM

**Video of image carousel in DataSession view**
https://github.com/LCOGT/datalab-ui/assets/54489472/7deb4c35-e359-4658-b2db-f3492e46a3ea

### DIFF
--- a/src/components/DataSession.vue
+++ b/src/components/DataSession.vue
@@ -1,8 +1,9 @@
 <script setup>
-// import { ref } from 'vue'
+import { ref, onMounted } from 'vue'
 import OperationPipeline from './OperationPipeline.vue';
 
 const emit = defineEmits(['reloadSession'])
+const images = ref([])
 
 const props = defineProps({
   data: {
@@ -26,6 +27,39 @@ function addOperation(operationDefinition) {
     .catch(error => console.error('Error:', error));
 }
 
+const getImages = async () => {
+  try {
+    const response = await fetch ('http://127.0.0.1:8000/api/datasessions/', {
+      method: 'GET',
+      headers: {
+        'Authorization': 'Token 123456789abcdefg',
+        'Content-Type': 'application/json; charset=UTF-8'
+      }
+    })
+
+    if (!response.ok) {
+      const errorData = await response.json()
+      console.error('Error Response Data:', errorData)
+      throw new Error('Bad request')
+    }
+
+    const data = await response.json()
+    console.log(data)
+    images.value = data.results
+    .flatMap(img => img.input_data)
+    // Filtering out items with 'archive' in the source
+    .filter(src => !src.source.includes('archive'))
+    .map(src => src.source)
+  } catch (error) {
+    console.log('Error getting images: ', error)
+  }
+
+}
+
+onMounted(() => {
+  getImages()
+})
+
 </script>
 
 <template>
@@ -35,11 +69,9 @@ function addOperation(operationDefinition) {
         <!-- TODO: Place input/output view here -->
         <v-carousel>
           <v-carousel-item
-            src="https://cdn.vuetifyjs.com/images/cards/sunshine.jpg"
-            cover
-          ></v-carousel-item>
-          <v-carousel-item
-            src="https://cdn.vuetifyjs.com/images/cards/docks.jpg"
+            v-for="(image, index) in images"
+            :key="index"
+            :src="require('../assets/' + image)"
             cover
           ></v-carousel-item>
         </v-carousel>
--- a/src/components/DataSession.vue
+++ b/src/components/DataSession.vue
@@ -44,7 +44,6 @@ const getImages = async () => {
     }
 
     const data = await response.json()
-    console.log(data)
     images.value = data.results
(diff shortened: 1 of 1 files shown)


src/components/DataSession.vue
```diff
@@ -26,6 +27,39 @@ function addOperation(operationDefinition) {
     .catch(error => console.error('Error:', error));
 }
 
+const getImages = async () => {
+  try {
+    const response = await fetch ('http://127.0.0.1:8000/api/datasessions/', {
```
mgdaily, line 32: I wonder if we can pass the data session ID down as a prop to this component so that we can make the API query specifically for the selected datasession?

src/components/DataSession.vue
```diff
@@ -26,6 +27,39 @@ function addOperation(operationDefinition) {
...
+      method: 'GET',
+      headers: {
+        'Authorization': 'Token 123456789abcdefg',
```
mgdaily, line 35: As we start to add more requests to the backend API, we may want to think about using ajax with a pre-filter so that we can set the auth headers for every request in a single place, just to avoid repeated code and having to re-define the headers for each request. Here's an example: https://github.com/observatorycontrolsystem/science-archive-client/blob/d557d535092a004d9e17ba79df313ce9941d2ac7/src/main.js#L42C7-L42C7

Not sure if this is supported in fetch - googling around, it doesn't seem like it is but we could always create a util function that wraps the fetch call and adds in the auth info automatically

src/components/DataSession.vue
```diff
@@ -26,6 +27,39 @@ function addOperation(operationDefinition) {
...
+    })
+
+    if (!response.ok) {
+      const errorData = await response.json()
+      console.error('Error Response Data:', errorData)
+      throw new Error('Bad request')
+    }
+
+    const data = await response.json()
+    console.log(data)
```
mgdaily, line 47: Do we need to keep the console.log of data in here right now? Seems useful for debugging but maybe not final code

LTDakin:
Im leaving it up to Matt to decide wether we should add this without this functionality "Another important note is that all of the images from the api are being displayed for each data session rather than only the images associated with that session. Trying to find a workaround, but likely, the workaround will be that we need to implement functionality for which session we want to add selectedImages to." Im not sure.

mgdaily (commented):
Nice! A few questions/comments before approval

src/components/DataSession.vue
capetillo: we can also install axios : )

src/components/DataSession.vue
capetillo: classic me leaving console logs

src/components/DataSession.vue
mgdaily: This article explains what I'm thinking about nicely - I think it'd be good to add an axios interceptor for requests to the backend!

https://blog.bitsrc.io/setting-up-axios-interceptors-for-all-http-calls-in-an-application-71bc2c636e4e

src/components/DataSession.vue
mgdaily: Or I guess somehow keep track of the selected data session in the store so we know which API resource we're looking for

src/components/DataSession.vue
capetillo: okey dokes! I'll read it 😎

capetillo: I'll make a story for this!

mgdaily: Sounds good!!

mgdaily (approved):

src/components/DataSession.vue
jnation3406: Sorry for responding late - the way I wrote the datasession is it gets all its data passed in through the data prop on creation - that should already contain the images and everything else it needs. You shouldn't need to do another get here, although we could change that for performance reasons in the future.

PR #20, opened 2024-01-09
Implemented Image List such that it uses store

LTDakin:
Implemented image list to use the vuex store so it would share a selected image array with the carousel. Now users can switch back and forth between the two different ways of viewing their images while keeping their selections. 

Future note: currently the selected images are not cleared when a project is switched. But maybe they should be kept to bring in images from multiple projects. Will need consideration. 

Changes: 
- added getters and setters to store to set entire array at once
- logic in image list to use store and load data from store on render
- linting auto changes for projectlist.js ( I'm not sure how to roll back but they follow the new rules anyways )
- removed unused-vars rule from eslint
- removed old logic that stored selectedImages as a variable in the parent of the carousel and list in favor of the store method

### DIFF
--- a/src/components/ProjectView/ImageList.vue
+++ b/src/components/ProjectView/ImageList.vue
@@ -19,16 +19,12 @@ let itemsPerPage = ref(15);
 let selected = ref([])
 
 // since item-value for v-data-table is set to basefilename we need to convert it back into an object
-// figuring out how to set item-value to the entire object would be great and then this function can be simplified
 function select(selectedImageNames) {
-  const selectedObjects = selectedImageNames.map(basefileName =>
-    items.value.find(item => item.basefile_name === basefileName)
-  );
-  store.dispatch('setSelectedImages', selectedObjects)
+  store.dispatch('setSelectedImages', selectedImageNames)
 }
 
 onMounted ( () => {
-  selected.value = store.getters.selectedImages.map(item => item.basefile_name)
+  selected.value = store.getters.selectedImages
 })
 </script>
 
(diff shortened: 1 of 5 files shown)


LTDakin:
Meeting with Jon at 11 on 1/9 to discuss implementation, then will request for review

src/components/ProjectView/ImageList.vue
```diff
@@ -1,41 +1,59 @@
...
+]);
+let items = ref(props.data);
+let itemsPerPage = ref(15);
+
+// --- Selection Logic ---
+let selected = ref([])
+
+// since item-value for v-data-table is set to basefilename we need to convert it back into an object
+// figuring out how to set item-value to the entire object would be great and then this function can be simplified
+function select(selectedImageNames) {
```
capetillo, line 23: you can use the `some` method! I don't quite understand the comment (because you've been working on the list view) but the `some` method checks if some elements of an array pass the test provided by the function. I think that's what you're trying to do, if not, I completely misunderstood.

LTDakin: Thats a really useful function to know I'll remember that! In this case the data table is just giving me back the basefile names of the selected objects since it uses them as a key to identify which are selected. So before storing them in the store I need to convert them back to their original image objects and not just the name.

src/components/ProjectView/ImageList.vue
```diff
@@ -1,41 +1,55 @@
...
+
+// v-data-table setup variables
+let headers = ref([
+  { title: "Image Name", align: "start", sortable: true, key: "basefile_name" },
+  { title: "Time", align: "start", sortable: true, key: "time" },
+  { title: "Object", align: "start", sortable: true, key: "object" },
+  { title: "Image", align: "start", sortable: false, key: "image" },
+]);
+let items = ref(props.data);
+let itemsPerPage = ref(15);
```
capetillo, line 16: I think we can change this to be an infinite scroll and add lazy loading. It can be a separate story though : )

capetillo:
i think it's okay to keep all selectedImages from different projects! It aligns with Wayne's views to let the kids experiment

capetillo (commented):
after the small changes, the code looks good to me!

src/components/ProjectView/ImageList.vue
LTDakin: added to board

capetillo (approved):
LGTM

PR #21, opened 2024-01-11
Feature/add images to datasession

capetillo:
### DESCRIPTION

**Background:**
The kids want to add images from their projects to existing data sessions or new data sessions. This is crucial for DataLab. Not sure I have much background on this one.

**Implementation:**
First, per Jon's suggestion, I got rid of the `ImportButton` component and refactored it in `ProjectView`. Here, there are 5 functions that allow for users to add to existing data sessions or create new ones. 

_getDataSessions_: gets the existing data session names from the api and then adding them to `uniqueDataSessions`. We also check for uniqueness to avoid displaying the same session names 

_addImagesToExistingSession_: used when the user selects images and wants to add them to an existing session. We make a PATCH request to the api to modify some of the data (Please let me know if the request body here should be modified. Perhaps get rid of 'name'?)

_selectDataSession_: removes the pop up that comes up when the user clicks the 'add to a session' button. It also triggers `addImagesToExistingSession` and reroutes user to `DataSessionView`

_createNewDataSession_: checks if new session name exists and if it does, returns an error text. If it doesn't, it makes a POST request to the api. It also resets state values to their default and redirects user to `DataSessionView`

_sessionNameExist_: pretty self explanatory. Used in the function above. 

Additionally, I added a `utils` folder with an `api.js` file that makes HTTP requests to the api to simplify the process of sending requests and handling responses by wrapping the native fetch api in a more convenient function. Thank you Matt for this suggestion.

### VISUALS

**Selecting an existing data session**

https://github.com/LCOGT/datalab-ui/assets/54489472/e4c655b6-6c72-4998-81e8-845b1e1eff6e

**Creating a new data session**

https://github.com/LCOGT/datalab-ui/assets/54489472/91cc793a-70e5-4bd5-b3c2-5d6bd29a069e

**Creating a new data session with an existing session name**

https://github.com/LCOGT/datalab-ui/assets/54489472/17a79ce2-8ad5-45c7-935d-c842777ef602


src/utils/api.js
```diff
@@ -0,0 +1,34 @@
...
+  
+      return await response.json()
+    } catch (error) {
+      console.error('Fetch API Call Error:', error)
+      throw error
+    }
+  }
+  
+  export { fetchApiCall }
+
```
mgdaily, line 34: Nice refactor here! just missing a newline

src/utils/api.js
```diff
@@ -0,0 +1,34 @@
...
+        }
+      }
+  
+      return await response.json()
+    } catch (error) {
+      console.error('Fetch API Call Error:', error)
+      throw error
+    }
+  }
+
```
mgdaily, line 32: I'm having a lot of trouble following the exception handling in this function. Having the entire function wrapped in a try-except block doesn't seem completely necessary. This is what I might do that I think accomplishes the same thing:

```javascript
async function fetchApiCall(url, method, body = null) {
      const headers = {
        'Content-Type': 'application/json',
        'Accept': 'application/json',
        'Authorization': 'Token 123456789abcdefg',
      }

      const config = {
        method: method,
        headers: headers,
        body: body ? JSON.stringify(body) : null
      }

      try {
        const response = await fetch(url, config)
      }
      catch(error) {
        console.error("Error raised when sending request.", error)
        throw Error("Error raised when sending request.", error)
      }

      if (!response.ok) {
        console.error("Response not OK", response)
        throw Error("Response not OK", response)
      }
      else {
        return await response.json()
      }
    }
  }
```

This way, we'll always throw a known error (our own) if fetch raises an exception, or the response in not OK (2XX). Then the caller can handle these as needed.

src/views/ProjectView.vue
```diff
@@ -1,9 +1,127 @@
...
+
+const router = useRouter()
+const store = useStore()
+const isPopupVisible = ref(false)
+const uniqueDataSessions = ref([])
+const newSessionName = ref('')
+const errorMessage = ref('')
+
+const getDataSessions = async () => {
+  try {
```
mgdaily, line 16: Do we need all of this in a try/except block? It's good practice to limit the scope of a try block to just the line or lines of code that may generate an exception - in this case, probably just the fetchApiCall() method.

src/views/ProjectView.vue
```diff
@@ -1,9 +1,127 @@
...
+      .map(session => ({ id: session.id, name: session.name }))
+
+    isPopupVisible.value = true
+  } catch (error) {
+    console.error('Error getting data sessions: ', error)
+  }
+}
+
+const addImagesToExistingSession = async (session) => {
+  try {
```
mgdaily, line 38: Limit the scope of this try block

src/views/ProjectView.vue
```diff
@@ -1,9 +1,127 @@
...
+        errorMessage.value = ''
+
+        router.push({ name: 'DataSessions' })
+    } catch (error) {
+        console.log('Error creating new data session:', error)
+        errorMessage.value = 'Error creating new data session'
+    }
+}
+
+const doesSessionNameExist = (name) => {
```
mgdaily, line 94: Minor, but `doesSessionNameExist` implies to me a 'yes/no' return value - what if we called it `sessionNameExists` so that it reflects that it's a boolean return? Naming things is hard

src/views/ProjectView.vue
```diff
@@ -1,9 +1,127 @@
...
+  router.push({ name: 'DataSessions' })
+}
+
+const createNewDataSession = async () => {
+    if (doesSessionNameExist(newSessionName.value)) {
+        errorMessage.value = 'Data Session name already exists. Please choose a different name.'
+        return
+    }
+
+    try {
```
mgdaily, line 70: Again, we probably want to limit the scope of this try block.

src/views/ProjectView.vue
```diff
@@ -1,9 +1,127 @@
...
+            <div v-if="errorMessage">{{ errorMessage }}</div>
+            </v-card-text>
+            <v-card-actions>
+            <v-btn color="primary" text @click="createNewDataSession">Create New Session</v-btn>
+            <v-btn color="primary" text @click="isPopupVisible = false">Close</v-btn>
+            </v-card-actions>
+        </v-card>
+    </v-dialog>
+  </template>
+
```
mgdaily, line 127: Missing newline.

src/utils/api.js
mgdaily: By limiting the scope of the try/except blocks we get more granular control of our exceptions and what exception we're passing the caller.

mgdaily (changes requested):
Mostly looking good - just some comments about the exception handling and naming

src/utils/api.js
capetillo: Isn't line 34 a new line?

src/utils/api.js
capetillo: thank you for your feedback! I'll make note of this for next time too : )

src/views/ProjectView.vue
capetillo: isn't line 127 the newline?

capetillo: am I misunderstanding something?

src/views/ProjectView.vue
capetillo: yes

src/utils/api.js
mgdaily: It's hard to say. POSIX standard requires a `\n` at the end of file (which VSCode can insert for you) for many unix tools to properly display/interpret these files. That blank line may be something different than the standard, and GitHub is showing that the file isn't properly terminated.

src/utils/api.js
mgdaily: Yes, sorry - I hastily refactored that into something that I thought made sense. Yours looks great

src/utils/api.js
capetillo: hmm on my file it does end with an extra line. Is that what it is? I can send you screenshots

src/utils/api.js
capetillo: sweet, thank you!

src/views/ProjectView.vue
mgdaily: You can read here - more specifically, GitHub is showing the file isn't properly terminated. https://www.baeldung.com/linux/files-end-with-newlines

capetillo: Screenshot 2024-01-11 at 9 32 31 AM
Screenshot 2024-01-11 at 9 32 55 AM

capetillo: github is gaslighting

src/views/ProjectView.vue
mgdaily: Was this updated?

mgdaily (changes requested):
I think we still need to limit the scope of the try clauses in a few of these functions

src/utils/api.js
```diff
@@ -1,34 +1,29 @@
 async function fetchApiCall(url, method, body = null) {
```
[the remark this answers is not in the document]
jnation3406: Here's an example of this type of functionality - notice the successCallback and failCallback that are passed into the function and then called in those cases. This helps us be able to use it within a lot of different components and situations. 

https://github.com/observatorycontrolsystem/ocs-component-lib/blob/main/src/mixins/getDataMixins.js#L40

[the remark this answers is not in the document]
jnation3406: I should note, that a mixin not a pure function so it expects those callbacks to be defined in the component you are mixing it into. You can also do it is a pure function though taking the function callbacks as parameters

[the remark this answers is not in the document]
capetillo: Solid advice. I can do that right now!

mgdaily (approved):
Thanks for making the changes!

PR #24, opened 2024-01-16
added a disabled condition to button props if no selected images

LTDakin:
added disabled prop to the add to session button to ensure the user has at least one image selected so they cant call the api to create empty data sessions or add 0 images to a data session.

### DIFF
--- a/src/views/ProjectView.vue
+++ b/src/views/ProjectView.vue
@@ -1,6 +1,6 @@
 <script setup>
 // libraries
-import { ref } from 'vue'
+import { ref, computed } from 'vue'
 // components
 import ProjectBar from '@/components/ProjectView/ProjectBar.vue';
 import ImageCarousel from '@/components/ProjectView/ImageCarousel.vue';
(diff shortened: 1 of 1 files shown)


src/views/ProjectView.vue
```diff
@@ -109,7 +109,7 @@ const sessionNameExists = (name) => {
             <ImageList v-if="!imageDisplayToggle" :data="MockData"/>
             <div class="control-buttons">
                 <v-switch class="d-flex mr-4" v-model="imageDisplayToggle" inset prepend-icon="mdi-view-list" append-icon="mdi-image"/>
-                <v-btn @click="getDataSessions">Add to a Session</v-btn>
+                <v-btn :disabled="store.getters.selectedImages.length === 0" @click="getDataSessions">Add to a Session</v-btn>
```
mgdaily, line 112: This seems like it might be cleaner as a computed property?

LTDakin: I can change that, I agree it was on the edge of pulling it out into a method or leaving in there.

mgdaily (approved):

PR #25, opened 2024-01-17
Fix image carousel display on DataSessions view + wrote handleError and handleSuccess callbacks 

capetillo:
### Description:

**Background:**

The images in DataSessions View were always showing an empty image (won't talk about that much in this PR)
And implementing best practices to be able to reuse fetchApiCall

**Implementation:**

For images, I just added an if statement to check for length.

Refactored `fetchApiCall` function (from `api.js`) by creating `handleSuccess` and `handleError` callback functions in `ProjectView`. Also, thanks to Matt's suggestion, I rewrote the way params are passed in `fetchApiCall` so that they're an object, that way it's easier to read and less room for error. 

_handleSuccess_: this callback manages successful api responses. It maps the results with keys of `id` and `name` and stores them in `uniqueDataSessions` 

_handleError_: this callback manages api call failures by logging errors

_addImagesToExistingSession_: Fixed patching here by using the spread operator and mapping the selected images as input data. Then sending that input data as part of the body. That way, instead of replacing the entire input data, it adds on to the existing one. 

I also got rid of the logic that filters through data session names and only adds one of each name, since that has to be handled in the backend and we want to display true data. 

### VISUALS

**Video of patching working as intended + images being displayed first thing when a data session is selected**

https://github.com/LCOGT/datalab-ui/assets/54489472/1c235ce9-a362-413e-9ea5-ffe2c7ca3832


src/views/ProjectView.vue
```diff
@@ -17,6 +17,7 @@ const isPopupVisible = ref(false)
 const uniqueDataSessions = ref([])
 const newSessionName = ref('')
 const errorMessage = ref('')
+const apiBaseUrl = 'http://127.0.0.1:8000/api/datasessions/'
```
mgdaily, line 20: Don't need to do this now, but we might want to create a story to put these endpoints in a centralized place, like PTR does - maybe from environment variables? :D 

https://github.com/LCOGT/ptr_ui/blob/dev/src/store/modules/api_endpoints.js

src/views/ProjectView.vue
```diff
@@ -26,85 +27,99 @@ const noSelectedImages = computed(() => {
   return store.getters.selectedImages.length === 0
 })
 
+// manages successful api response by mapping data to unique sessions
+const handleSuccess = (data) => {
```
mgdaily, line 31: Now that I think about this more, maybe we can rename this function to what it's doing since we're only using it for a single API call? handleSuccess isn't immediately descriptive

src/views/ProjectView.vue
```diff
@@ -26,85 +27,99 @@ const noSelectedImages = computed(() => {
...
+    try {
+        // fetches existing session data
+        const currentSessionResponse = await fetchApiCall({ url: sessionIdUrl, method: 'GET' })
+        const currentSessionData = currentSessionResponse.input_data
 
-    const requestBody = {
-      'name': session.name,
-      'input_data': inputData
-    }
+        // merging existing and new image data
```
mgdaily, line 61: Maybe add a comment with a TODO that this is temporary until we change it in the backend?

mgdaily (commented):
Mostly looks good! Nice job on the refactor. Just a few comments

src/views/ProjectView.vue
capetillo: 100% yes

mgdaily (approved):

PR #26, opened 2024-01-19
Feature/delete data sessions

LTDakin:
New Features
- Sessions can now be deleted through the delete dialog, the page then reloads to show this change occurred
Changes
- broke off Delete Dialog into its own component for organization
- shortened name of ref for keeping track of the displayed boolean
Fixes
- added handling to the fetch api utility so that it won't cause an error when receiving empty responses
- minor spaces and lines

### DIFF
--- a/src/components/DeleteSessionDialog.vue
+++ b/src/components/DeleteSessionDialog.vue
@@ -1,0 +1,54 @@
+<script setup>
+import { fetchApiCall } from '@/utils/api';
+const props = defineProps([ 'modelValue', 'deleteId'])
+const emit = defineEmits(['update:modelValue', 'refreshDataSessions'])
+
+function closeDialog() { emit('update:modelValue', false)}
+
+async function confirmDeleteSession() {
+  if (props.deleteId != -1) {
+    console.log('deleting session with id ' + props.deleteId)
+    const url = 'http://127.0.0.1:8000/api/datasessions/'+props.deleteId+'/'
+    try {
+      await fetchApiCall(url, 'DELETE' )
+      .then(response => response.json())
+    } catch (error) {
+      console.log('error calling api: ', error)
+    }
+    emit('refreshDataSessions')
+    closeDialog()
+  }
+}
+</script>
+<template>
+  <!-- Shared dialog used to confirm deleting of sessions -->
+  <v-dialog
+    :model-value="modelValue"
+    persistent
+    width="auto"
+  >
+    <v-card>
+      <v-card-title class="text-h5" color="red-lighten-1">
+        Delete Session
+      </v-card-title>
+      <v-card-text>Are you sure you want to delete this Datalab Session? This operation is not reversible!</v-card-text>
+      <v-card-actions>
+        <v-spacer></v-spacer>
+        <v-btn
+          color="green-darken-1"
+          variant="text"
+          @click="closeDialog()"
+        >
+          Go Back
+        </v-btn>
+        <v-btn
+          color="red-darken-1"
+          variant="text"
+          @click="confirmDeleteSession()"
+        >
+          Delete
+        </v-btn>
+      </v-card-actions>
+    </v-card>
+  </v-dialog>
+</template>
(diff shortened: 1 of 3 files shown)


src/components/DeleteSessionDialog.vue
```diff
@@ -0,0 +1,50 @@
+<script setup>
+import { fetchApiCall } from '@/utils/api';
+const props = defineProps([ 'modelValue', 'deleteId'])
+const emit = defineEmits(['update:modelValue', 'reloadSession'])
+
+function closeDialog() { emit('update:modelValue', false)}
+
+async function confirmDeleteSession() {
+  if (props.deleteId != -1) {
+    const url = 'http://127.0.0.1:8000/api/datasessions/' + props.deleteId
```
mgdaily, line 10: I think I mentioned this in Carolina's PR, but we probably want to make a story to get these URLs into a central place (like the store) and read in from environment variables

src/components/DeleteSessionDialog.vue
```diff
@@ -0,0 +1,50 @@
...
+    fetchApiCall({url:url, method:'DELETE'})
```
mgdaily, line 11: @capetillo's fetchApiCall can take callbacks on success/fail, so we should probably do these callbacks that way to stay consistent

src/utils/api.js
```diff
@@ -14,7 +14,16 @@ async function fetchApiCall({ url, method, body = null, successCallback = null,
 
   try {
     const response = await fetch(url, config)
+
+    // Check for empty or non-JSON response
+    if (response.status === 204 || response.status === 301) {
```
mgdaily, line 19: Hmm, this seems oddly specific. Why do we need to handle these specific codes this way? Is there a better way than checking status codes?

mgdaily (commented):
A few comments/questions

src/utils/api.js
jnation3406: This seems like some boilerplate/copied code... I don't think we will get either of those codes from our server response

LTDakin: `[19/Jan/2024 00:09:27] "DELETE /api/datasessions/23 HTTP/1.1" 301 0
[19/Jan/2024 00:09:27] "DELETE /api/datasessions/23/ HTTP/1.1" 204 0`

The api sends these two responses from the server on a Delete request. Both which have empty responses that the fetchAPI method can't handle currently. But, I think a better way would be checking the response body itself for an empty one and handling that instead of the specific codes. Before it would send an error whenever I sent a DELETE

src/components/DeleteSessionDialog.vue
```diff
@@ -0,0 +1,57 @@
...
+
+let showSnackBar = ref(false)
+
+function closeDialog() { 
+  emit('update:modelValue', false)
+  emit('reloadSession')
+}
+
+async function confirmDeleteSession() {
+  const url = 'http://127.0.0.1:8000/api/datasessions/' + props.deleteId
```
mgdaily, line 15: Let's update this to use carolina's central config that she merged to main?

mgdaily (approved):

capetillo (commented):
add spaces in params

src/views/DataSessionsView.vue
```diff
@@ -25,14 +26,6 @@ function loadAllSessions() {
     .then(data => dataSessions.value = data.results);
```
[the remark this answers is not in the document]
LTDakin: let me write a different story for refactoring Jon's code. We need to include using the callbacks for api calls as well as using the store for the url. "Refactoring Jon's API calls" or something 😅

src/components/DeleteSessionDialog.vue
```diff
@@ -0,0 +1,50 @@
+<script setup>
+import { fetchApiCall } from '@/utils/api';
+const props = defineProps([ 'modelValue', 'deleteId'])
+const emit = defineEmits(['update:modelValue', 'reloadSession'])
+
+function closeDialog() { emit('update:modelValue', false)}
+
+async function confirmDeleteSession() {
```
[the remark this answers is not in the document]
LTDakin: I think you might have commented on an older commit. The newest one has await and async

mgdaily (approved):
Looks good!

PR #27, opened 2024-01-19
Feature/centralize items

capetillo:
Won't write a long PR for this one. Not feeling great.
Added a public > config > config.json
Here, I added a json object literal with just a few key/value pairs (for now). I wrote an async function in api.js called `loadConfig()` where we return the config data to be used in App.vue. I also added a few consts to the store and initialized them as empty strings. The consts are api urls and isConfigLoaded which was initialized to false.
In App.vue, I added `onMounted` and here we call `loadConfig`. If we get config data, then we change isConfigLoaded to true, and mutate the state to the config data url values


public/config/config.json
```diff
@@ -0,0 +1,4 @@
+{
+    "dataSessionApiUrl": "http://127.0.0.1:8000/api",
+    "dataLabARchiveApiUrl": "https://datalab-archive.photonranch.org/"
```
jnation3406, line 3: ARchive

src/components/RegistrationView/TestingUser.vue
```diff
@@ -0,0 +1,92 @@
...
+                            </v-card-text>
+                        </v-card>
+                    </v-tab-item>
+                </v-tabs>
+            </div>
+        </v-dialog>
+    </v-app>
+</div>
+    <button @click="loginAndGetToken">Login and Get Token</button>
+  </template>
```
jnation3406: the spacing looks off here?

src/utils/api.js
```diff
@@ -1,4 +1,18 @@
-// handles api requests with configurable parameters and callback functions
+
+async function loadConfig() {
```
jnation3406, line 2: This is something that I would call once on Vue app load, and place the values in the store. Then different components can just get what they need from the store, i.e. getBackendUrl or whatever. That way you aren't making the fetch of the file more than one time if its needed in multiple places.

src/views/ProjectView.vue
```diff
@@ -122,9 +124,18 @@ const sessionNameExists = (name) => {
...
+  apiBaseUrl = config.dataSessionApiUrl + '/datasessions/'
+  // for rendering purposes
+  isConfigLoaded.value = true
+})
+
 </script>
 <template>
-    <div class="container">
+    <!-- only load if config is loaded -->
+    <div v-if="isConfigLoaded" class="container">
```
jnation3406, line 138: If you load the config file on Vue App load, you will not need to do this here or in any other components

jnation3406 (changes requested):
Looks good, Just a recommendation to move where the loadConfig is called to before Vue App initialization and store results in the store - then the values can be used in any component without having to fetch again, and you will have them ready before any components load.

src/components/RegistrationView/TestingUser.vue
capetillo: dang I meant to delete this file! not relevant to this pr, sorry

src/utils/api.js
capetillo: I see! Okey dokes, gonna do that!

public/config/config.json
capetillo: it's for quirky purposes

src/App.vue
```diff
@@ -1,8 +1,27 @@
...
+import { onMounted, computed } from 'vue'
+import { useStore } from 'vuex'
+import { loadConfig } from './utils/api'
+
+const store = useStore()
+const loadedConfig = computed(() => store.state.isConfigLoaded)
+
+// invoking loadConfig when app first mounts
+onMounted(async () => {
+  const config = await loadConfig()
```
mgdaily, line 11: nitpick - we're probably only going to use loadConfig here right? if so I would say it doesn't necessarily need to be a util function. especially in the util/api section, since it's not calling an API. if it's specific to loading config on app initialization you could just put the logic inline here.

capetillo: aah okey dokes!

capetillo: fixed!

PR #28, opened 2024-01-26
Feature/registration functionality

capetillo:
### DESCRIPTION
**Background:**
Although we can't really implement log in functionality at the moment, there is a fake super user with mock data from the OCS. 

**Implementation:**
I created a view called RegistrationView with log in functionality. The order of operations is: 1. get token 2. use token to get user's profile 3. save user's profile and projects to the store

_authenticateAndGetToken_: makes a POST request to get token and stores token in store and sends what the user writes in `username` and `password` fields as credentials. 

_getUser_: uses the token to make a GET request to get the user's profile and if successful, saves the profile and projects to the store and reroutes user to ProjectView

Added rules to make username and password required fields.
I also made a few changes in the config file and therefore had to implement changes elsewhere (ProjectView and DeleteSessionDialog) but functionality is the same.

### VISUALS
**Log in functionality when credentials are not that of super user**

https://github.com/LCOGT/datalab-ui/assets/54489472/bc122c18-e69d-423c-a478-8e63ec5a3a55

No video if log in functionality works because token and username are displayed


src/views/RegistrationView.vue
```diff
@@ -0,0 +1,108 @@
...
+
+async function authenticateAndGetToken() {
+    const authUrl = profileBaseUrl + 'api-token-auth/'
+    const credentials = {
+      username: username.value,
+      password: password.value
+  }
+
+  try {
+    const response = await fetch(authUrl, {
```
LTDakin, line 29: Should we make a function in the api.js file for these requests to keep it consistent? 
with this modification the function can take an optional token parameter. 

Did not check how the OCS frontend handles it as it uses a passthrough component, maybe Matt or Jon have another api util file example with a different solution for handling multiple apis with and without tokens.

```
async function fetchApiCall({ url, method, body = null, successCallback = null, failCallback = null, token=null }) {
  const headers = {
    'Content-Type': 'application/json',
    'Accept': 'application/json',
// this line needs some changing to only include the key Authorization if theres a token
    'Authorization': token ? token : 'Token 123456789abcdefg', 
  }
```

public/config/config.json
```diff
@@ -1,5 +1,5 @@
 {
   "apiBaseUrl": "http://127.0.0.1:8000/api/",
-  "dataSessionsUrl": "http://127.0.0.1:8000/api/datasessions/",
-  "dataLabArchiveUrl": "https://datalab-archive.photonranch.org/"
+  "dataLabArchiveApiUrl": "https://datalab-archive.photonranch.org/",
+  "ocsApiUrl": "https://observe.photonranch.org/api/"
```
mgdaily, line 4: We might want to make this observationPortalApiUrl - OCS is a set of services, and we're specifically targeting the observation portal here

src/views/RegistrationView.vue
```diff
@@ -0,0 +1,108 @@
+<script setup>
+import { ref } from 'vue'
+import { useStore } from 'vuex'
+import { useRouter } from 'vue-router'
+
+const store = useStore()
+const router = useRouter()
+const profileBaseUrl = store.state.ocsApiUrl
```
mgdaily, line 8: The naming here could potentially be confusing - yeah, we're getting the profile information from the observation portal but an unsuspecting person might be confused as to why it's called profileBaseUrl here. You could get around this by just using the store.state value directly when you build the URL rather than assigning it to a separate variable here.

src/views/RegistrationView.vue
mgdaily: You could also add an optional headers param, so it's more flexible than just a token!

src/views/ProjectView.vue
```diff
@@ -17,7 +17,7 @@ const isPopupVisible = ref(false)
 const uniqueDataSessions = ref([])
 const newSessionName = ref('')
 const errorMessage = ref('')
-const dataSessionsUrl = store.state.dataSessionsUrl
+const apiBaseUrl = store.state.apiBaseUrl
```
mgdaily, line 20: I'm losing track of the different APIs we're talking to - we may want to change this to datalabApiBase or something similar in the store. Consistency in naming is going to pay off big time here!

src/store/index.js
```diff
@@ -33,12 +37,30 @@ export default createStore({
...
 
-    setDataSessionsUrl(state, url) {
-      state.dataSessionsUrl = url
-    },
-
     setArchiveUrl(state, url) {
       state.archiveUrl = url
+    },
+    
+    setOcsApiUrlUrl(state, url) {
```
mgdaily, line 44: The double "url" here, while I get why you did it, seems like it could be confusing

mgdaily (commented):
Some comments, mostly on naming! Naming is hard.

src/views/RegistrationView.vue
```diff
@@ -0,0 +1,108 @@
+<script setup>
+import { ref } from 'vue'
+import { useStore } from 'vuex'
+import { useRouter } from 'vue-router'
+
+const store = useStore()
+const router = useRouter()
+const profileBaseUrl = store.state.userProfileUrl
```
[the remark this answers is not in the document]
capetillo: I created the variable cause I had to append to it 😬

src/views/RegistrationView.vue
```diff
@@ -0,0 +1,108 @@
...
+}
+
+async function authenticateAndGetToken() {
+    const authUrl = profileBaseUrl + 'api-token-auth/'
+    const credentials = {
+      username: username.value,
+      password: password.value
+  }
+
+  try {
```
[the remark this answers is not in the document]
capetillo: thank you, it took me 7 tries to learn

src/views/ProjectView.vue
capetillo: lol

capetillo: yes, naming is hard

src/store/index.js
capetillo: that was an accident oopsiiiii

src/views/RegistrationView.vue
LTDakin: Matt's idea is more customizable so I agree

src/components/DeleteSessionDialog.vue
```diff
@@ -7,8 +7,15 @@ const props = defineProps([ 'modelValue', 'deleteId'])
 const emit = defineEmits(['update:modelValue', 'reloadSession'])
 const store = useStore()
 
+const authHeaders = {
+    'Content-Type': 'application/json',
+    'Accept': 'application/json',
+    'Authorization': 'Token 123456789abcdefg',
```
jnation3406, line 13: Shouldn't this get the authtoken from the store now?

src/views/RegistrationView.vue
```diff
@@ -0,0 +1,105 @@
...
+    router.push({ name: 'ProjectView' })
+}
+
+const getUserProfile = async () => {
+    await authenticateUser ()
+    const token = store.state.authToken
+    const authHeaders = {
+        'Content-Type': 'application/json',
+        'Accept': 'application/json',
+        'Authorization': token ? `Token ${token}` : null,
```
jnation3406, line 64: I think this is unneccessary because token will either be empty string or a value you set it to, so either way you will fail authentication on the request if its empty string or just empty...

src/views/ProjectView.vue
```diff
@@ -17,7 +17,13 @@ const isPopupVisible = ref(false)
 const uniqueDataSessions = ref([])
 const newSessionName = ref('')
 const errorMessage = ref('')
-const dataSessionsUrl = store.state.dataSessionsUrl
+const dataSessionsUrl = store.state.datalabApiBaseUrl + 'datasessions/'
+
+const authHeaders = {
+    'Content-Type': 'application/json',
+    'Accept': 'application/json',
+    'Authorization': 'Token 123456789abcdefg',
```
jnation3406, line 25: shouldn't this also use the authtoken from the store?

jnation3406 (approved):
Looks good just a few comments on switching out the hardcoded api tokens.

src/components/DeleteSessionDialog.vue
capetillo: noted, used the one from the store now!

src/views/RegistrationView.vue
capetillo: true, updated!

src/views/ProjectView.vue
capetillo: done!

mgdaily (approved):

PR #30, opened 2024-01-29
Feature/fetch hosted mock data

LTDakin:
New
- Function loadUserImages that fetches all frames with an authtoken from the archive, called on component mount and stores frames in a ref named UserFrames
- ref userFrames that holds the user's images from the archive
- A skeleton loader that is displayed when we don't have the image data on hand yet

Changed
- updating property names so that images will be working from the new format instead of the mock data format
- pointing the datalab api url to the server link instead of a locally running instance
- removed the filter and map step when displaying data session images so that the key and alt text can be the file name instead of an index for future image operation selection.

Fixes
- missing image alt text
- binding of transition to a number with v-bind instead of treating the 0 as a string which caused warnings

Side note: we have updated the datalab url to point to the server api instead of a local one. However the original implementation didn't use the fetchApi util and so Data Session View is non functional. A fix updating all of those api requests to use the store url and api util function will fix this.

Screenshot 2024-01-29 at 2 15 43 PM

### DIFF
--- a/src/views/ProjectView.vue
+++ b/src/views/ProjectView.vue
@@ -14,16 +14,18 @@ const uniqueDataSessions = ref([])
 const newSessionName = ref('')
 const errorMessage = ref('')
 const dataSessionsUrl = store.state.dataSessionsUrl
-const data = ref(null)
 let imageDisplayToggle = ref(true)
+let userFrames = ref(null)
 
-// Logic for fetching data from OCS and Archive for displaying in Projects
-const getUserImages = async () => {
-    // url and body for getting the list of the user's images
-    const url = ""
-    const body = ""
-    const imageLinks = await fetchApiCall(url, 'GET', body, console.log("success"), console.log("failure"))
-    data.value = await fetchImagesfromLinks(imageLinks)
+// Loads the user's Images from their profile into userImages
+const loadUserImages = async () => {
+    // TODO Get profile Data from store
+    // Get s3 image links for user for all proposals
+    // user links to get images from the archive
+    // cache images a ref
+    const url = 'https://datalab-archive.photonranch.org/frames/?reduction_level=95'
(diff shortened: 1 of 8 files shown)


src/views/ProjectView.vue
```diff
@@ -27,7 +24,13 @@ const authHeaders = {
 
 // toggle for optional data viewing, controlled by a v-switch
 let imageDisplayToggle = ref(true)
+let userFrames = ref(null)
 
+// Loads the user's Images from their profile into userImages ( currently just fetches all frames from archive regardless of proposal )
+const loadUserImages = async () => {
+    const url = archiveUrl + 'frames/?reduction_level=95'
+    userFrames.value = await fetchApiCall({url: url, method: 'GET', token: 'Token 2da74cb5262a52bc479e5b63b548fd5910592475'})
```
capetillo, line 32: use `store.state.authToken`

capetillo (commented):
hopefully you can read my comment

src/views/ProjectView.vue
LTDakin: oops! good catch

src/views/ProjectView.vue
```diff
@@ -27,7 +24,13 @@ const authHeaders = {
 
 // toggle for optional data viewing, controlled by a v-switch
 let imageDisplayToggle = ref(true)
+let userFrames = ref(null)
 
+// Loads the user's Images from their profile into userImages ( currently just fetches all frames from archive regardless of proposal )
+const loadUserImages = async () => {
```
mgdaily, line 30: Suggestion: We could put the reduction level as an argument to this function so that it's more reusable

mgdaily (approved):
One suggestion/comment. Otherwise looks good!

PR #32, opened 2024-01-30
Fix/update datasession api fetches

LTDakin:
Putting up a working PR but could use some improvements, since this fixes the site temporarily however I think it should be added and the improvements added on another PR. Open to changing them in this one, but won't be ready for the demo.

When the url for the datalab api was changed in the config one PR ago we didn't change the old fetch requests that were still pointing towards the locally hosted api. Updated plain fetch requests with the util function and pointed them to the server api.

Improvements/Tech Debt for a follow up story:

- initially followed the way it was done and copied auth headers for every file that needs a token request. Should be moved to a shared file or handled in the api util. This way the headers parameter can be eliminated from most calls, but still have the option to have a custom header if needed
 - n̶o̶t̶ ̶u̶s̶i̶n̶g̶ ̶t̶h̶e̶ ̶s̶u̶c̶c̶e̶s̶s̶ ̶c̶a̶l̶l̶b̶a̶c̶k̶s̶,̶ ̶w̶a̶n̶t̶e̶d̶ ̶t̶o̶ ̶t̶h̶i̶n̶k̶ ̶m̶o̶r̶e̶ ̶o̶n̶ ̶a̶ ̶c̶l̶e̶a̶n̶ ̶w̶a̶y̶ ̶t̶o̶ ̶i̶m̶p̶l̶e̶m̶e̶n̶t̶ ̶c̶a̶l̶l̶b̶a̶c̶k̶ ̶f̶u̶n̶c̶t̶i̶o̶n̶s̶ ̶t̶o̶ ̶d̶o̶ ̶t̶h̶e̶ ̶s̶m̶a̶l̶l̶ ̶t̶a̶s̶k̶s̶ ̶r̶e̶q̶u̶i̶r̶e̶d̶ ̶b̶y̶ ̶m̶o̶s̶t̶ ̶c̶a̶l̶l̶s̶ ̶
- added the handle error to the api util function that just prints out the error to console, but errors should be handled on a case by case basis to allow elegant failure

### DIFF
--- a/src/components/DataSession.vue
+++ b/src/components/DataSession.vue
@@ -1,5 +1,5 @@
 <script setup>
-import { ref, onMounted } from 'vue'
+import { ref, onMounted, defineEmits, defineProps } from 'vue'
 import OperationPipeline from './OperationPipeline.vue'
 import { fetchApiCall, handleError } from '../utils/api'
 import { useStore } from 'vuex'
--- a/src/components/DataSession.vue
+++ b/src/components/DataSession.vue
@@ -29,7 +29,7 @@ async function addOperation(operationDefinition) {
 
 const getImages = async () => {
 	const url = dataSessionsUrl + props.data.id
-	const data = await fetchApiCall({url: url, method: 'GET', headers: authHeaders, successCallback: emit('reloadSession'), failCallback: handleError})
+	const data = await fetchApiCall({url: url, method: 'GET', headers: authHeaders, failCallback: handleError})
 	images.value = data.input_data
 }
 
--- a/src/components/DataSession.vue
+++ b/src/components/DataSession.vue
@@ -29,8 +29,7 @@ async function addOperation(operationDefinition) {
 
 const getImages = async () => {
 	const url = dataSessionsUrl + props.data.id
-	const data = await fetchApiCall({url: url, method: 'GET', headers: authHeaders, failCallback: handleError})
-	images.value = data.input_data
+	await fetchApiCall({url: url, method: 'GET', headers: authHeaders, successCallback: (data) => {images.value = data.input_data}, failCallback: handleError})
 }
 
 onMounted(() => {
--- a/src/components/DataSession.vue
+++ b/src/components/DataSession.vue
@@ -1,9 +1,19 @@
 <script setup>
 import { ref, onMounted } from 'vue'
 import OperationPipeline from './OperationPipeline.vue'
+import { fetchApiCall, handleError } from '../utils/api'
+import { useStore } from 'vuex'
 
+const store = useStore()
 const emit = defineEmits(['reloadSession'])
 let images = ref([])
+const dataSessionsUrl = store.state.datalabApiBaseUrl + 'datasessions/'
+
+const authHeaders = {
+	'Content-Type': 'application/json',
+	'Accept': 'application/json',
+	'Authorization': `Token ${store.state.authToken}`,
+}
 
 const props = defineProps({
 	data: {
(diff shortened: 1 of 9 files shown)


src/components/DataSession.vue
```diff
@@ -12,43 +22,15 @@ const props = defineProps({
...
-			'Authorization': 'Token 123456789abcdefg',
-			'Content-type': 'application/json; charset=UTF-8'
-		}
-	})
-		.then(response => response.json())
-		.then(function() {emit('reloadSession')})
-		.catch(error => console.error('Error:', error))
+async function addOperation(operationDefinition) {
+	const url = dataSessionsUrl + props.data.id + '/operations/'
+	await fetchApiCall({url: url, method: 'POST', body: operationDefinition, headers: authHeaders, successCallback: emit('reloadSession'), failCallback: handleError})
```
capetillo, line 27: I _think_ we have to `JSON.stringify(operationDefinition)`

src/components/DataSession.vue
```diff
@@ -12,43 +22,15 @@ const props = defineProps({
...
-		}
-
-		const data = await response.json()
-		images.value = data.input_data
-	} catch (error) {
-		console.log('Error getting images: ', error)
-	}
-
+	const url = dataSessionsUrl + props.data.id
+	const data = await fetchApiCall({url: url, method: 'GET', headers: authHeaders, failCallback: handleError})
```
capetillo, line 32: I think the `successCallback` here could be something as simple as `images.value = data.input_data`

src/components/DataSession.vue
LTDakin: I tested my changes to make sure they work before putting up any PRs. It actually caught me by surprise cause in an earlier commit I didn't change it and I got an error. But it turns out the fetchapi handles it already with this -> 'body: body ? JSON.stringify(body) : null'

src/components/OperationWizard.vue
```diff
@@ -1,25 +1,28 @@
...
-		.then(response => response.json())
-		.then(data => {
-			availableOperations.value = data
-			if (Object.keys(availableOperations.value).length > 0){
-				selectedOperation.value = Object.keys(availableOperations.value)[0]
-			}
-		})
+onMounted(async () => {
+  const url = dataSessionsUrl + 'available_operations/'
+  availableOperations.value = await fetchApiCall({url: url, method: 'GET', headers: authHeaders, failCallback: handleError})
```
capetillo, line 22: I think this should be in a try/catch block and the successCallback could handle assigning the `selectedOperation.value` to `availableOperations`

src/views/DataSessionsView.vue
```diff
@@ -17,13 +27,9 @@ function deleteSession(id) {
...
 
-function loadAllSessions() {
-	// Load in the session details from the API
-	fetch('http://127.0.0.1:8000/api/datasessions/', {
-		'headers': {'Authorization': 'Token 123456789abcdefg'}
-	})
-		.then(response => response.json())
-		.then(data => dataSessions.value = data.results)
+async function loadAllSessions() {
+  const data = await fetchApiCall({url: dataSessionsUrl, method: 'GET', headers: authHeaders, failCallback: handleError})
```
capetillo, line 31: sorry for the same comment, but again I think this needs to be in a try/catch block and `datasessions.value = data.results` should be the handlesuccess callback

capetillo (commented):
Great work! Some comments/suggestions

src/components/DataSession.vue
LTDakin: I thought passing an assignment as a function argument is a big no no, but I might be wrong. could make a function that just has that as a one liner. Wasn't sure the best way to implement these so I'd be open to discussion.

src/views/DataSessionsView.vue
mgdaily: @capetillo your comment got me to go back and check the implementation - is there any instance fetchApiCall willraise an error? Looking at the implementation it looks like it handles all exceptions and will not raise any to the caller, so is a try/catch really needed on the fetchApiCall?

mgdaily: But I do agree that the one-liner success callbacks are generally fine and you can easily put those inline in the call to fetchApiCall, so we should probably stick to that until we decide if it needs a refactor or not.

LTDakin: haven't used call backs much but if its ok then ill write it in as a lambda

src/components/OperationWizard.vue
LTDakin: same comment for the rest in terms of the successcallback.  except want to call about the try catch since I thought fetchAPI already catches its own errors

mgdaily: See my previous comment, but I am not sure fetchApiCall can raise an error that needs to be caught?

src/components/DataSession.vue
mgdaily: I'd just define it as an inline arrow function for now - happy to talk about a refactor later on if we need to.

src/components/DataSession.vue
capetillo: oh yeah! good eye

src/views/DataSessionsView.vue
capetillo: @mgdaily it isn't necessary, it's just customary in async functions. But yeah, `fetchApiCall` already has a try/catch. My PRs that also use `fetchApiCall` got approved with try/catch so I thought we were implementing what's generally used to avoid going up the chain to handle rejection

mgdaily: > @mgdaily it isn't necessary, it's just customary in async functions. But yeah, `fetchApiCall` already has a try/catch. My PRs that also use `fetchApiCall` got approved with try/catch so I thought we were implementing what's generally used to avoid going up the chain to handle rejection

yeah, I'm just having this realization now. they were approved, but I hadn't realized we weren't capable of passing any exceptions back up to the caller. I don't know if there's a JS best practice, but in general if a function can't actually raise an exception then a try/catch block is not strictly necessary

mgdaily (approved):
Most of my comments were similar to Carolina's. Looks good

src/views/DataSessionsView.vue
capetillo: it's not necessarily best practice. From what I know, it's just to stop going up the chain but i guess failcallback does that. I WILL NOT LET THIS FEED MY IMPOSTER SYNDROME

PR #33, opened 2024-01-31
Feature/display large images

capetillo:
### Description

**Background**
We want the functionality for when a user selects a thumbnail to display its corresponding large image. The current implementation is not necessarily the long term one, but it works with the amount of data that we currently have. 

**Implementation**
When Project View first mounts, we call `loadLargeImages` which then uses the util api call. The success callback awaits for the `data.results`, which are objects with the images' metadata, and then these get stored in the store through `setLargeImages` mutation. I also have a `firstLargeImage` in the store which is to hardcode the first image so that when the user first loads the Project View, they can see the first large image. 

Now, here's the logic on how to display the selected thumbnail as a large image. First, in `handleThumbnailClick` we assign the value of `lastSelectedImage` 's basename (minus `-small`) to `smallImageBasename`. Then, in `getLargeImageSource` we access the store to get the array of objects of `largeImages` and find one whose basename (minus `-large`) matches that on `smallImageBaseName` and we get the url from that object. That url is then assigned to `largeImageSrc`.

In the template, the image source checks if there is a `largeImageSrc`. If there is, that's the source, if there isn't (i.e. the user just loaded the Project View page, then we get `firstLargeImage` from the store 

### Visuals
**If user is logged in, first image to load is the first one from the array. You can see that as user selects other thumbnail images, the corresponding large image is displayed**

https://github.com/LCOGT/datalab-ui/assets/54489472/98a417f4-cd41-42bc-a2dc-a29024631c6c

### DIFF
--- a/src/components/ProjectView/ImageCarousel.vue
+++ b/src/components/ProjectView/ImageCarousel.vue
@@ -16,16 +16,20 @@ let data = ref(props.data)
 // Checking if image isSelected to either add or remove yellow borderline
 const isSelected = (item) => store.getters.isSelected(item)
 
-const getLargeImageBasename = () => {
-	const largeImages = store.state.largeImages
-	// loading first large image for user to see when they first navigate the page
-	if (!smallImageBasename.value) {
-		largeImageSrc.value = largeImages[0].url
-	} else {
-		let selectedLargeImage = largeImages.find(obj => obj.basename.replace('-large', '') === smallImageBasename.value)
-		largeImageSrc.value = selectedLargeImage.url
+const getLargeImageSource = () => {
+	if (store.state.largeImages) {
+		const largeImages = store.state.largeImages
+		console.log('largeimages:', largeImages[0].url)
+		// loading first large image for user to see when they first navigate the page
+		if (!smallImageBasename.value) {
+			console.log('we are here')
+			largeImageSrc.value = largeImages[0].url
+		} else {
+			let selectedLargeImage = largeImages.find(obj => obj.basename.replace('-large', '') === smallImageBasename.value)
+			largeImageSrc.value = selectedLargeImage.url
+		}	
 	}
-	
+
 }
 // Invoked any time an image is clicked
 const handleThumbnailClick = (item, index) => {
@@ -39,11 +43,11 @@ const handleThumbnailClick = (item, index) => {
 	} else {
 		currentSlide.value = index
 	}
-	getLargeImageBasename()
+	getLargeImageSource()
 }
 
 onMounted(() => {
-	handleThumbnailClick()
+	getLargeImageSource()
 })
 
 </script>
--- a/src/components/ProjectView/ImageCarousel.vue
+++ b/src/components/ProjectView/ImageCarousel.vue
@@ -108,6 +108,10 @@ const getImageSrc = (src) => {
 </template>
 
 <style scoped>
+
+.carousel__item {
+  padding-bottom: 2em;
+}
 .thumbnail__item {
   max-width: 200px;
   max-height: 160px;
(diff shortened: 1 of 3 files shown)


src/store/index.js
```diff
@@ -11,7 +11,9 @@ export default createStore({
 			username: '',
 			authToken: '',
 			profile: [],
-			projects: []
+			projects: [],
+			largeImages: [],
+			firstLargeImage: ''
```
mgdaily, line 16: would it make sense for the firstLargeImage just be a getter on largeImages? e.g.

```
  getters: {
    firstLargeImage (state) {
      return state.largeImages[0].url
    }
  }
```

https://vuex.vuejs.org/guide/getters.html#getters

src/views/ProjectView.vue
```diff
@@ -131,8 +131,20 @@ const sessionNameExists = (name) => {
...
+	store.commit('setLargeImages', largeImages)
+	store.commit('setFirstLargeImage', largeImages[0].url)
+}
+
+const loadLargeImages = async (option) => {
+	const loadLargeImageUrl = option ? archiveUrl + 'frames/?' + option : archiveUrl + 'frames/'
+	await fetchApiCall({url: loadLargeImageUrl, method: 'GET', headers: authHeaders, successCallback: saveLargeImages, failCallback: handleError})
+}
+
 onMounted(() => {
```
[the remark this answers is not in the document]
capetillo: forgot about the spinning feature!

src/store/index.js
capetillo: done!

src/store/index.js
```diff
@@ -61,6 +63,16 @@ export default createStore({
...
+		},
+
+		setLargeImages(state, images) {
+			for (const image of images) {
+				state.largeImages.push(image)
+			}
+		},
+
+		setFirstLargeImage(state, image) {
+			state.firstLargeImage =image
```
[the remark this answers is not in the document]
capetillo: deleted this, so no longer needed : )

src/components/ProjectView/ImageCarousel.vue
```diff
@@ -53,7 +57,7 @@ const handleThumbnailClick = (item, index) => {
     >
       <div class="carousel__item">
         <img
-          :src="item.url"
+          :src="largeImageSrc ? largeImageSrc : store.state.firstLargeImage"
```
LTDakin: related to my previous comment about the getter but if we have a getter then this can just be `:src="store.getters.getLargeImageSrc(smallImageBasename)"` and we don't need the ref or function `getLargeImageSource()`

LTDakin (commented):
Only one I think is needed is my comment about setLargeImages. As for the getter to simplify the code and leave the logic of getting the large image to store thats something to think about whether it makes sense or not.

src/store/index.js
```diff
@@ -61,6 +63,16 @@ export default createStore({
 			for (const project of projects) {
 				state.projects.push(project)
 			}
+		},
+
+		setLargeImages(state, images) {
+			for (const image of images) {
```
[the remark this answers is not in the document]
capetillo: good eye

src/views/ProjectView.vue
```diff
@@ -131,8 +131,19 @@ const sessionNameExists = (name) => {
 	return uniqueDataSessions.value.some(session => session.name === name)
 }
 
+const saveLargeImages = async (data) => {
+	const largeImages = await data.results
+	store.commit('setLargeImages', largeImages)
+}
+
+const loadLargeImages = async (option) => {
```
jnation3406, line 139: Just an idea - we could have the store maintain the logic for retrieving a large image from source. I.e. move this function into the store as an action it can do. Then have a function in the store that is get a large (or small) image given its basename. The store would look in its cache, and if not present retrieve it an put it in the cache. This way multiple components (such as the wizard, datasession view, and project view) could all just call the same store function to get the images they want to display, and the code to retrieve and cache those images would be reused within the store.

mgdaily (approved):
Approved! Pending addressing Lloyd's comment about getLargeImageSource

PR #34, opened 2024-02-01
Update/persist login credentials

LTDakin:
The site will now keep people logged in between page refreshes. It does this by storing the auth token in local storage through the vuex store and the package persisted state.

Changes:
- added npm package vuex-persistedstate
- added store module userData that is persisted and contains user profile and auth token
- added default header to fetchApi util and removed declarations of auth headers it replaces


jnation3406 (approved):
Looks good to me

PR #37, opened 2024-02-07
login page fixes, css, and default page

LTDakin:
Some changes to the registration page of datalab. I'm trying out a new way of writing these PRs. Let me know if breaking it up by file makes it easier to review.

**Index.html**
- Set app to take up the full screen

**App.vue**
- used template for logical display since it was adding an empty div previously

**Index.js**
- Login is now the landing page
- Deleted HomeView component

**RegistrationView.vue**
- removed try catch from util function since it handles its own errors
- removed auth header since util function handles it now
- added autocomplete attributes so browsers can save user and pass
- styled login page for tablets

All views
- check for auth token and routing to login if not present, or routing from login to projectview if it is present

Screenshot 2024-02-06 at 4 10 23 PM


src/App.vue
```diff
@@ -27,9 +27,9 @@ onMounted(async () => {
 </script>
 
 <template>
-  <div v-if="loadedConfig">
+  <template v-if="loadedConfig">
```
capetillo, line 30: out of curiosity, why are there nested templates?

src/views/RegistrationView.vue
```diff
@@ -30,78 +31,95 @@ const handleError = (error) => {
...
-		handleError(error)
-	}
-}
-
 const storeUser = (user) => {
 	store.dispatch('userData/setUsername', username.value)
 	store.dispatch('userData/setUserProfile', user)
 	store.commit('setProjects', user.proposals)
-	router.push({ name: 'ProjectView' })
+	router.push({ name: 'DataSessions' })
```
capetillo, line 38: I think users should be rerouted to projects (since Wayne says that that's the first thing that users should see)

capetillo (commented):
one curious question and one suggestion for rerouting user to Projects after logging in : ) other than that, great job!

src/App.vue
LTDakin: I just learned this today from this question, nice little trivia -> https://stackoverflow.com/questions/51995815/vuejs-template-vs-div-or-related-for-grouping-elements-for-conditional-rend

src/views/RegistrationView.vue
LTDakin: Did this cause the project view doesn't have a button to go to data sessions so to get off the screen you have to create a session. But actually for now I can revert this and we can add a button to go to Data sessions on the project view in the case you don't want to make a new session.

src/router/index.js
```diff
@@ -1,14 +1,13 @@
...
-import Home from '../views/HomeView.vue'
 import Registration from '../views/RegistrationView.vue'
 
 const routes = [
 	{
 		path: '/',
-		name: 'Home',
-		component: Home
+		name: 'Registration',
+		component: Registration
```
mgdaily, line 10: If we navigate to the root / page and we're already logged in (e.g. auth data exists in store) does this redirect to the appropriate place, e.g. projects page?

mgdaily (commented):
Just a question since you changed the root page to the registration page - I don't see logic in there to redirect if the user's already logged in.

We could use a navigation guard for that purpose if we want - and we could have a getter in the store that determines if the user's logged in based on what's in the store.

And breaking it up by file definitely helps!

src/router/index.js
LTDakin: I didn't implement either of these since I wasn't sure if we'd be combining this with photon ranch in the future. But that would be so far out now that I think about it and the logic for rerouting shouldn't take too long. I'll add that now.

src/views/DataSessionsView.vue
```diff
@@ -14,6 +14,10 @@ const deleteSessionId = ref(-1)
 const showDeleteDialog = ref(false)
 const dataSessionsUrl = store.state.datalabApiBaseUrl + 'datasessions/'
 
+onBeforeMount(()=>{
+	if(!store.getters['userData/authToken']) router.push({ name: 'Registration' })
```
mgdaily, line 18: Only comment I have is that you could make a getter like `userIsAuthenticated` in the userData portion of the store that checks if the authToken exists. It's mostly syntactic sugar, though

mgdaily (approved):

PR #41, opened 2024-02-12
Styling/datasession

capetillo:
### Description
**Background:**
After discussing in our design meeting, we agreed that images should be displayed in a grid system rather than a carousel

**Implementation:**
In the `Datasession` component, I added a `calculateColumnSpan` function to determine the number of columns in the grid system. This function calculates the number of columns each image should span to fit up to 4 images (subject to change) in a row on a 12 column grid (vuetify's limit, but we can change this later), without exceeding the total number of images. If there are fewer than 4 images, it adjusts so that the images can evenly spread across the grid. 
Important note: I left some commented out code cause there's a bug that has been fixed on a separate branch. The commented out code displays the `progress-circular` component from vuetify to display when images are loading. This commented out code will soon not be commented out (check next PR)

### Visuals
**Data Session grid view on desktop**
Screenshot 2024-02-12 at 11 54 41 AM

**Data Session grid views on iPad**
1F8A8E84-3313-4D47-936E-6FADE2B714ED
D1E6F297-C28F-4414-ABCA-8D113FE15D93

### DIFF
--- a/src/components/DataSession.vue
+++ b/src/components/DataSession.vue
@@ -28,13 +28,10 @@ async function addOperation(operationDefinition) {
 	await fetchApiCall({url: url, method: 'POST', body: operationDefinition, successCallback: emit('reloadSession'), failCallback: handleError})
 }
 
-const displayImages = (data) => {
-	images.value = data.input_data
-}
 
 const getImages = async () => {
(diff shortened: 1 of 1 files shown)


mgdaily (approved):
Lookin' good!

PR #42, opened 2024-02-12
Fix/add images to datasession

capetillo:
### Description
**Background:**
There was a bug where we couldn't add images to existing data sessions because the structure of the keys from images objects changed

**Implementation:**
Fairly simple implementation. In `addImagesToExistingSession` found in Project View, I changed the `source` and `basename` under `inputData` to `image.url` and `image.basename`, respectively, because we were sending empty objects. Now we can add images to existing data sessions : )

### DIFF
--- a/src/components/DataSession.vue
+++ b/src/components/DataSession.vue
@@ -28,14 +28,9 @@ async function addOperation(operationDefinition) {
 	await fetchApiCall({url: url, method: 'POST', body: operationDefinition, successCallback: emit('reloadSession'), failCallback: handleError})
 }
 
-const displayImages = (data) => {
-	images.value = data.input_data
-	console.log('images.value', images.value)
-}
-
 const getImages = async () => {
 	const url = dataSessionsUrl + props.data.id
-	await fetchApiCall({url: url, method: 'GET', successCallback: displayImages, failCallback: handleError})
+	await fetchApiCall({url: url, method: 'GET', successCallback: (data) => { images.value = data.input_data }, failCallback: handleError})
 }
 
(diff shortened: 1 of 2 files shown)


src/components/DataSession.vue
```diff
@@ -39,31 +50,46 @@ onMounted(() => {
...
-          :operations="data.operations"
-          @add-operation="addOperation"
-        />
+        <v-img
+          :src="image.source"
+          :alt="image.basename"
+          cover
+          aspect-ratio="1"
+        >
+          <!-- <template #placeholder>
```
mgdaily, line 68: Commented code needed?

src/components/DataSession.vue
```diff
@@ -28,9 +28,20 @@ async function addOperation(operationDefinition) {
 	await fetchApiCall({url: url, method: 'POST', body: operationDefinition, successCallback: emit('reloadSession'), failCallback: handleError})
 }
 
+const displayImages = (data) => {
+	images.value = data.input_data
+	console.log('images.value', images.value)
```
mgdaily, line 33: Extraneous console log? I think having this inline like it was before is just fine if it's a one-liner

capetillo: oopsi

src/components/DataSession.vue
capetillo: deleted!

mgdaily:
Looks like there are some merge conflicts that need to be resolved!

mgdaily (approved):
lookit that small diff! nice work

PR #43, opened 2024-02-14
Chore/basename to backend

capetillo:
### Description
**Background:**
Jon brought to our attention that we should only send the basename to the backend and the `source` should be where to point at.

**Implementation:**
In `ProjectView` where the user can create a new data session or add to a data session, I changed the source and the basename. The basename also gets rid of `-small` or `-large` to send the true basename to the backend. 
In the `DataSession` component, we use the  props that are passed down from `DataSessionView` and loop through the `inputData`. From here, we make an api request to get the image urls by using the basename and appending `-small` since we currently only need the small images. 
I also changed the successCallback to push the `data.results` to `images.value`

### Visuals

**Video of everything working as normal : )**

https://github.com/LCOGT/datalab-ui/assets/54489472/3dccd2e8-3adf-4908-926a-4f8ae4c88ee6

### DIFF
--- a/src/components/DataSession.vue
+++ b/src/components/DataSession.vue
@@ -4,18 +4,17 @@ import OperationPipeline from './OperationPipeline.vue'
 import { fetchApiCall, handleError } from '../utils/api'
 import { useStore } from 'vuex'
 
+const store = useStore()
+const emit = defineEmits(['reloadSession'])
+let images = ref([])
+const dataSessionsUrl = store.state.datalabApiBaseUrl + 'datasessions/'
+
 const props = defineProps({
 	data: {
 		type: Object,
 		required: true
 	}
 })
-
-const store = useStore()
-const emit = defineEmits(['reloadSession'])
-let images = ref([])
-const dataSessionsUrl = store.state.datalabApiBaseUrl + 'datasessions/'
-
 
 async function addOperation(operationDefinition) {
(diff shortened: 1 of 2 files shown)


src/components/DataSession.vue
```diff
@@ -45,18 +50,17 @@ onMounted(() => {
...
 
-
 <template>
   <v-container class="d-lg-flex">
     <v-row v-if="images.length">
       <v-col
-        v-for="image in images"
-        :key="image.basename"
+        v-for="image of images"
+        :key="image[0].basename"
```
mgdaily, line 58: Is there a reason we need to do this indexing of 0? Is images an array of objects, so each object just has a basename and a url? Or am I missing something

src/views/ProjectView.vue
```diff
@@ -82,10 +82,11 @@ const addImagesToExistingSession = async (session) => {
 }
 
 // closes popup, invokes addImagesToExistingSession, and reroutes user to DataSessions view
-const selectDataSession = (session) => {
+const selectDataSession = async (session) => {
 	isPopupVisible.value = false
-	addImagesToExistingSession(session)
+	await addImagesToExistingSession(session)
```
mgdaily, line 87: pesky await

mgdaily (commented):
Mostly looks good! Nice small PR - just trying to understand one bit about the images array

mgdaily (approved):

PR #45, opened 2024-02-20
Feature/wizard image selection

capetillo:
### Description
**Background:**
The Wizard wasn't displaying images before. It's important to display images and be able to select them to know what images we want to manipulate.

**Implementation:**
Started by passing down `images` as props to be able to display all of the images related to a data session in the wizard. Then I added basically the same selection implementation from image carousel here. I also implemented the same function as in the Data Session view to calculate column span. A _**separate**_ story could be globalizing some of these functions/components. 
The next step was to add the selected images as input files to be able to post them. So I did that like this:
```
'input_data': {
	...selectedOperationInput.value,
	'input_files': selectedDataSessionImages.value
}
```
and then I got rid of this line in the `addOperation` function
```
operationDefinition.input_data.input_files = structuredClone(images.value)
```

### Visuals
**Screenshot of the api response in `addOperation` having an image as the `input_files`** 
Screenshot 2024-02-20 at 3 14 11 PM

**Video of The Wizard flow**

https://github.com/LCOGT/datalab-ui/assets/54489472/810dbbed-a3fb-4d2a-a3d3-6c59503190bc

### DIFF
--- a/src/components/OperationWizard.vue
+++ b/src/components/OperationWizard.vue
@@ -73,7 +73,7 @@ const goForwardText = computed(() => {
 		return 'Configure Operation'
 	}
 	else {
-		return 'Add Operation'
+		return 'Add Operationzzz'
 	}
 })
 
@@ -95,7 +95,10 @@ function goForward() {
 		displayImages.value = false
 		let operationDefinition = {
 			'name': selectedOperation.value,
-			'input_data': selectedOperationInput.value
+			'input_data': {
+				...selectedOperationInput.value,
+				'input_files': selectedDataSessionImages.value
+			}
 		}
 		emit('addOperation', operationDefinition)
 	}
--- a/src/components/OperationWizard.vue
+++ b/src/components/OperationWizard.vue
@@ -235,6 +235,7 @@ const handleThumbnailClick = (item) => {
   max-width: 20%; 
   height: auto; 
   box-sizing: border-box; 
+  cursor: pointer;
 }
 .selected-image {
   border: 0.3rem solid $dark-green;
--- a/src/components/OperationWizard.vue
+++ b/src/components/OperationWizard.vue
@@ -10,6 +10,7 @@ const dataSessionsUrl = store.state.datalabApiBaseUrl
 const availableOperations = ref({})
 const selectedOperation = ref('')
 const selectedOperationInput = ref({})
+const selectedDSImages = ref([])
 
 let displayImages = ref(false)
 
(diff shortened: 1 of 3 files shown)


src/components/OperationWizard.vue
```diff
@@ -145,6 +179,26 @@ function goForward() {
         </v-row>
       </v-card-text>
     </v-slide-y-reverse-transition>
+    <v-row
```
jnation3406, line 182: Line 171 should be where this goes - i.e. we should render an image picker when there is an input option of type `file`. Then you can remove the placeholder text I have there instead of the widget

src/components/OperationWizard.vue
```diff
@@ -78,16 +89,36 @@ const wizardTitle = computed(() => {
...
+
+const handleThumbnailClick = (item) => {
+	const index = selectedDataSessionImages.value.findIndex(selectedImage => selectedImage.basename === item.basename)
+	if (index === -1) {
+		selectedDataSessionImages.value.push(item)
+	} else {
+		selectedDataSessionImages.value.splice(index, 1)
+	}
+}
+
```
jnation3406, line 121: I think the picker code should be its own vue component. We already do something really similar in the project image selection view, and I suspect there will be other places we may want to use it in the future as well. It should be pretty self contained of show a bunch of image thumbnails, allowing to scroll through them and select them, and emit the selection...

jnation3406 (changes requested):

src/components/OperationWizard.vue
capetillo: I agree AND this should be a separate story 😬

src/components/OperationWizard.vue
capetillo: a fun challenge. I'll work on it

src/components/OperationWizard.vue
```diff
@@ -125,26 +157,45 @@ function goForward() {
...
             :label="inputDescription.name"
             :type="inputDescription.type"
+            class="operation-input"
           />
-        </v-row>
+          <div
+            v-else-if="inputDescription.type == 'file'"
+            class="images-container"
+          >
+            <v-col
```
jnation3406, line 179: I didn't know v-col worked without being inside a v-row

jnation3406 (approved):
Looks good!

src/components/OperationWizard.vue
capetillo: the more you know

PR #46, opened 2024-02-22
Feature/analyis component

LTDakin:
This PR adds the analysis component to Datalab. It will take an image as a prop and then display it and allow for zooming, dragging, and downloading. To get this to work in image carousel the method for getting large images had to be overhauled.

Changes by File

ImageAnalyzer.vue
- new component for analyzing images
- uses component zoomableImage

ImageCarousel.vue
- removed prop 'data' which represented small images since we store them in the vuex store anyways
- renamed and set small image and large image at component creation
- deleted method getLargeImageSource and created it as a getter in store (previously requested)
- cleaned up logic of handleThumbnailClick using new getters
- replaced prop use of data in html with store

ImageList.vue
- removed prop data since we keep a list of the small images in the store

ZoomableImage.vue
- new component that displays an image in a container sized for the image and allows for dragging and zooming
- the way this component works is when a user clicks we save that point as mouseX, mouseY and attach an event listener for pointer move which calculates the delta for mouse movement. This delta is then applied to the image's top and left positioning using vue3 style binding.
- currently the image is bounded by 800 px but in a future story would like to bound it instead by checking whether overflow is occurring.

store/index.js
- removed mutation that could set two states to two separate mutations for best practice. 
- instead added the logic check for option to the action to determine which mutation to call
- removed firstLargeImage getter as it was unused
- added getter largeImage which can take a basename and return the corresponding large image
- NOTE currently we only load the caches from projectview, in the future we might want to load it from app.vue since it won't exist for datasessions view.
- fixed bug where components would try to load when there were no images by changing a check for null to be a check for length, added check for largeimagecache as well.

### DIFF
--- a/src/components/ImageAnalyzer.vue
+++ b/src/components/ImageAnalyzer.vue
@@ -1,12 +1,13 @@
 <script setup>
 import { defineProps, defineEmits } from 'vue'
+import ZoomableImage from './ZoomableImage.vue'
 
+// eslint-disable-next-line no-unused-vars
 const props = defineProps(['modelValue', 'image'])
 const emit = defineEmits(['update:modelValue'])
 
 function closeDialog() { 
 	emit('update:modelValue', false)
-	console.log(props.image)
 }
 
 </script>
@@ -18,6 +19,7 @@ function closeDialog() {
     <v-card class="pa-6">
       <v-toolbar>
         <v-spacer />
+        <h3>{{ image.target_name }}</h3>
         <a
           :href="image.url"
           :download="image.basename"
@@ -33,10 +35,7 @@ function closeDialog() {
           @click="closeDialog()"
         />
       </v-toolbar>
-      <img
-        class="image"
-        :src="image.url"
-      >
+      <zoomable-image :image-src="image.url" />
     </v-card>
   </v-dialog>
 </template>
--- a/src/components/ImageAnalyzer.vue
+++ b/src/components/ImageAnalyzer.vue
@@ -1,0 +1,32 @@
+<script setup>
+import { defineProps, defineEmits } from 'vue'
+
+const props = defineProps(['modelValue', 'image'])
+const emit = defineEmits(['update:modelValue'])
+
+function closeDialog() { 
+	console.log('image prop:', props.image)
+	emit('update:modelValue', false)
+}
+
+</script>
+<template>
+  <v-dialog
+    :model-value="modelValue"
+    fullscreen
+  >
+    <v-card>
+      <v-img
+        :src="hello"
+        cover
+      />
+      <v-btn
+        color="green-darken-1"
+        variant="text"
+        @click="closeDialog()"
+      >
+        Close
+      </v-btn>
+    </v-card>
+  </v-dialog>
+</template>
--- a/src/components/ImageAnalyzer.vue
+++ b/src/components/ImageAnalyzer.vue
@@ -36,7 +36,10 @@ function closeDialog() {
         />
       </v-toolbar>
       <div class="image_area">
-        <zoomable-image :image-src="image.url" />
+        <zoomable-image
+          v-if="image"
+          :image-src="image.url"
+        />
       </div>
     </v-card>
   </v-dialog>
--- a/src/components/ImageAnalyzer.vue
+++ b/src/components/ImageAnalyzer.vue
@@ -35,12 +35,15 @@ function closeDialog() {
           @click="closeDialog()"
         />
       </v-toolbar>
-      <zoomable-image :image-src="image.url" />
+      <div class="image_area">
+        <zoomable-image :image-src="image.url" />
+      </div>
     </v-card>
   </v-dialog>
 </template>
 <style scoped>
-.image{
-	object-fit: contain
+.image_area{
+	display: flex;
+	justify-content: center;
 }
 </style>
--- a/src/components/ImageAnalyzer.vue
+++ b/src/components/ImageAnalyzer.vue
@@ -16,15 +16,16 @@ function closeDialog() {
     :model-value="modelValue"
     fullscreen
   >
-    <v-card class="pa-6">
-      <v-toolbar>
-        <v-spacer />
-        <h3>{{ image.target_name }}</h3>
+    <v-card>
+      <v-toolbar
+        density="comfortable"
+        :elevation="8"
+        :title="image.target_name"
+      >
         <a
           :href="image.url"
           :download="image.basename"
           target="_blank"
-          style="text-decoration:none"
         >
           <v-btn
             icon="mdi-download"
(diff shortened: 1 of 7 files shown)


src/components/ImageAnalyzer.vue
```diff
@@ -0,0 +1,61 @@
...
+        <v-btn
+          icon="mdi-close"
+          @click="closeDialog()"
+        />
+      </v-toolbar>
+      <div class="anaylzer_content">
+        <zoomable-image
+          :image-src="image.url"
+        />
+        <v-sheet
```
LTDakin, line 43: this v sheet is the right hand side that contains a few details. Was created from Jon's input that we should have space on the right for some outputs

src/components/ProjectView/ImageCarousel.vue
```diff
@@ -1,47 +1,28 @@
...
-let smallImageBasename = ref('')
-let largeImageSrc = ref('')
-
-let data = ref(props.data)
-
-// Checking if image isSelected to either add or remove yellow borderline
-const isSelected = (item) => store.getters.isSelected(item)
-
-// Getting large images by using the thumbnail's basename and finding the large image that matches it
-const getLargeImageSource = () => {
```
LTDakin, line 21: this method became a getter in the store since that is the source of truth for our images, at least for the project view

src/components/ZoomableImage.vue
```diff
@@ -0,0 +1,90 @@
...
+const imageLeftPos = ref(0)
+const imageTopPos = ref(0)
+const currentZoom = ref(1)
+
+const MIN_ZOOM = 1
+const MAX_ZOOM = 3
+const STEP_SIZE = 0.1
+let mouseX
+let mouseY
+
```
LTDakin, line 15: Zoom will be run whenever the magnify buttons are pressed, mouse wheel is scrolled, or fingers pinched. It takes an optional arg which is direction and then uses the stepsize to change the images scale accordingly. Overflow on the image's container is hidden which gives it a zoom effect

src/components/ZoomableImage.vue
```diff
@@ -0,0 +1,90 @@
...
+const zoom = (e, direction) => {
+	if (!direction) direction = e.deltaY > 0 ? -1 : 1
+	let newZoom = currentZoom.value + direction * STEP_SIZE 
+	if (newZoom > MIN_ZOOM && newZoom < MAX_ZOOM) currentZoom.value = newZoom
+}
+
+// save the position clicked and add the pointermove listener
```
LTDakin, line 22: when the user clicks on the image we want to start listening for a mouse movement event

src/components/ZoomableImage.vue
```diff
@@ -0,0 +1,90 @@
...
+const pointerDown = (e) => {
+	mouseX = e.pageX
+	mouseY = e.pageY
+	image.value.onpointermove = pointerMove
+}
+
+// use mouse movement delta to set position, limit so image stays on screen
```
LTDakin, line 29: this function is what actually moves the image. ImageLeftPos and imageTopPos are bound css style variables. We are finding the delta from the last mouse position to the new one and applying that to the css styling which allows us to move the image around

src/components/ZoomableImage.vue
```diff
@@ -0,0 +1,90 @@
...
+    class="image_container"
+    :style="{height: image?.naturalHeight, width: image?.naturalWidth}"
+  >
+    <img
+      ref="image"
+      :src="imageSrc"
+      class="image"
+      :style="{left:imageLeftPos+'px', top:imageTopPos+'px', transform: `scale(${currentZoom})`}"
+      @wheel="zoom"
+      @pointerdown.prevent="pointerDown"
```
LTDakin, line 52: when the mouse is released or the mouse leaves the image area we want to stop listening for move events aka stop moving the image around

src/store/index.js
```diff
@@ -99,9 +93,16 @@ export default createStore({
 
 		selectedImages: (state) => state.selectedImages,
 
-		firstLargeImage: (state) => {
-			if (state.largeImageCache.length) {
-				return state.largeImageCache[0].url
+		// given an image basename, fetches the large res of that image
+		largeImage: (state) => (imageBasename) => {
```
capetillo, line 97: What happens when the Project View page first loads? I ask because I see that in `handleThumbnailClick` this getter is being used. But `handleThumbnailClick` is only invoked when clicking thumbnails

src/components/ZoomableImage.vue
```diff
@@ -0,0 +1,90 @@
...
+// save the position clicked and add the pointermove listener
+const pointerDown = (e) => {
+	mouseX = e.pageX
+	mouseY = e.pageY
+	image.value.onpointermove = pointerMove
+}
+
+// use mouse movement delta to set position, limit so image stays on screen
+const pointerMove = (e) => {
+	let x 						= imageLeftPos.value+e.pageX-mouseX
```
capetillo, line 31: not sue if this is just github being weird but I see something like this:
`let x                                                            = imageLeftPos. etc`
and same for everything in this function

src/store/index.js
```diff
@@ -58,18 +58,11 @@ export default createStore({
...
-				let imageArray
-				image.RLEVEL === 95 ? imageArray = state.smallImageCache : imageArray = state.largeImageCache
-				const existingImageIndex = imageArray.findIndex(img => img.basename === image.basename)
-				if (existingImageIndex === -1) {
-					imageArray.push(image)
-				} else {
-					imageArray[existingImageIndex] = image
-				}
-			})
+		setSmallImageCache(state, imageData){
```
capetillo, line 61: My humble opinion is that we should keep the setImageCache function as it was because:
1. it can be used for both large and small images
2. it checks to see if the image already exists in the array so we make less api requests

capetillo (commented):
some comments, questions, and suggestions

src/store/index.js
LTDakin: I think a mutation should only mutate one state and any decision making between which mutation to call should be done in an action.
As for the logic in the old function it was actually doing extra work. At this point we've already made the api call and gotten the list of images so it doesn't make less api requests. The backend is the source of truth. So we don't even need to check for duplicates and just replace the entire array.

src/store/index.js
LTDakin: at line 12 in project view when the component is created

src/components/ZoomableImage.vue
LTDakin: I did that to line up all the math and make it more readable, but I can change that if its not helpful then

src/store/index.js
LTDakin: @jnation3406 @mgdaily Carolina and I sat and talked about this and this is how she originally did it too, but someone said to put the logic for mutating both states into one mutation. Would like to hear your input on this if possible!

mgdaily (changes requested):
My main issue is the added coupling of the ImageCarousel and ImageList components to the store for their data, rather than having them read in from a prop passed in from the parent.

src/components/ProjectView/ImageCarousel.vue
```diff
@@ -114,7 +96,7 @@ const getImageSrc = (src) => {
     class="thumbnail__carousel__container"
   >
     <div
-      v-for="(item, index) in data"
+      v-for="(item, index) in store.state.smallImageCache"
```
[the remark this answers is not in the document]
LTDakin: oops I didn't think about that and agreed. I thought it would be cleaner to not use a prop if we already had the data in store but it makes more sense for the future to use the store in the parent component and pass it down to the children.

jnation3406 (approved):
Looks pretty good to me. We will need to have a discussion on the image cache - I think its a pretty "dumb" cache right now with the contents getting reset on each call to the api. Longer term I think well need to make it smarter, i.e. like a LRU cache with some max size, and then just keep adding things to it until its full or the users session expires. Because ideally we can cache the thumbnails in your browser over multiple page reloads over multiple days/weeks as you go between many projects / datasessions with different images and add more and more new images and so on.

mgdaily (approved):

PR #49, opened 2024-02-27
Chore/globalize components

capetillo:
## CHORE: Globalize Image Grid Component 

_**Important note before reading**_
You may see that many files have changed, but that's because I created a couple of new folders and rearranged files under the components folder. Here are the three files to review: `ImageGrid`, `OperationWizard`, and `DataSession`. The rest had either spacing changes, renaming of props (from images to data), or rearranged the order of files. 

**Background:**
The ImagerGrid component is used in Data Session to display the images related to the session and in the Operation Wizard to select the images that will be operated on. Because this component is used in more than one place, it has now been globalized. 

**Implementation**
`ImageGrid` : a lot of existing logic is now here. This includes how we get images and store them locally. 
`data` in `defineProps` comes from `DataSession` and it's an object with the images' metadata. `columnSpan` is used to calculate the column span to display the image grid. `selectedImages` is used to toggle the selection of images in the operation wizard and the function `isSelected` works with this to toggle selection.

I also changed the name of props in DataSession, OperationPipeline, and OperationWizard because it was getting confusing.  All of it is named `data` now. 

### VISUALS
Here are screenshots of everything working as normal

Screenshot 2024-02-27 at 9 09 26 AM

Screenshot 2024-02-27 at 9 10 15 AM


src/components/DataSession/DataSession.vue
capetillo, line 1: The major change here is the removal of logic and importing of imageGrid

src/components/DataSession/OperationWizard.vue
```diff
@@ -176,21 +172,13 @@ const handleThumbnailClick = (item) => {
...
-              <v-img
-                :src="image.url"
-                :alt="image.basename"
-                cover
-                aspect-ratio="1"
-                :class="{ 'selected-image': selectedDataSessionImages.some(selectedImage => selectedImage.basename === image.basename) }"
-                @click="handleThumbnailClick(image)"
-              />
-            </v-col>
+              @image-clicked="handleThumbnailClick"
```
capetillo, line 180: added this emit because imagegrid is a whole row so if we clicked on an image, the whole row would get a green border

src/utils/common.js
capetillo, line 1: Not sure why this is coming up as a new file but you just reviewed this so kindly, ignore : )

LTDakin (commented):
Looks good! Just one comment about a magic number.

src/components/DataSession/DataSession.vue
```diff
@@ -27,54 +28,23 @@ async function addOperation(operationDefinition) {
...
-          :src="image.url"
-          :alt="image.basename"
-          cover
-          aspect-ratio="1"
-        />
-      </v-col>
-    </v-row>
+    <image-grid
+      :data="data"
+      :column-span="calculateColumnSpan(data.input_data.length, 4)"
```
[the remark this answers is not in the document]
capetillo: magic number?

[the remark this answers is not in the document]
LTDakin: The magic number in this case is 4, its magic because its not easily apparent to a reader of the code what 4 is representing. This SO explains them in more detail. 
https://stackoverflow.com/questions/47882/what-are-magic-numbers-and-why-do-some-consider-them-bad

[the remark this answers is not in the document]
capetillo: Done! Thank you : ) I added both magic numbers in both cases

src/store/index.js
```diff
@@ -58,7 +58,7 @@ export default createStore({
 				state.projects.push(project)
 			}
 		},
-		setSmallImageCache(state, imageData){
+		setSmallImageCache(state, imageData) {
```
mgdaily, line 61: cute whitespace

mgdaily (approved):
Looks good! Onto the next PR!

PR #50, opened 2024-02-27
Feature/colorblind friendly

capetillo:
## FEATURE: Colorblind friendly

**Background:**
We want to be as inclusive as possible. The first step is to add the option of high contrast between text and background. 

**Implementation:**
I had to change the language from scss to css so that I can dynamically change variables and append `colorblind=true` to the entire document. For this reason, you can ignore most file changes. The changes were all related to removing `lang=scss` and changing the way colors were used. 
These are the files that you can review: `App.vue`, `store>index.js`, `ColorblindToggle.vue` and if you want to take a look at the colors, `custom-colors.css`
The logic is simple: I added a state to the store where `isColorblindMode` is set to false. The state is mutated to true if the user clicks on the `ColorblindToggle`. I also added a watcher to this state and when it is true, it sets an attribute of `colorblind` and it's removed when it's false. In `custom-colors.css` I added a `[colorblind="true"]` set of colors and a `:root` set of colors. That's all : ) 

### VISUALS
**Quick screen recording of the changes**

https://github.com/LCOGT/datalab-ui/assets/54489472/9272ce25-50b4-4802-8815-d61d40c1b767


jnation3406 (approved):
This is giving me deja-vu to when I first looked at your prototype and it had the toggle for dark-mode or not. And then I removed the toggle and made it only dark mode. And now you've added a color-blind toggle that makes it even darker than dark mode... and its tempting me

PR #53, opened 2024-02-29
Chore/display proper data

capetillo:
## CHORE: Display the correct large image when user unselects images

**Background:**
If user navigated to Project X and selected 2 images, then navigated to Project Y and selected 1 image and then unselected it, the large image displayed would be the last one selected from Project X. This is now fixed

**Implementation:**
In `handleThumbnailClick`, I added a couple of checks. 
First, `lastSelectedImages` checks if the `selectedImages` from store has length. If so, then we return the last one added to the array. Otherwise, we return null.
Then, `isLastSelectedImageInCurrentProject` has an existence check for `lastSelectedImage` (`&&` is an existence check in JS). And then it checks if some image in `props.data` has a basename that's equal to that of `lastSelectedImage`. 
If all of this is true, then we proceed as usual. Else, we set the `currentSlide` value to be that of index. 

The end

### Visuals
**Screen recording of the navigation of this process with the correct images displayed**

https://github.com/LCOGT/datalab-ui/assets/54489472/41bcc1fa-8f8c-44d7-aba6-6db49a93126f

### DIFF
--- a/src/components/Project/ImageCarousel.vue
+++ b/src/components/Project/ImageCarousel.vue
@@ -8,10 +8,14 @@ import ImageAnalyzer from '../ImageAnalyzer.vue'
 
 const props = defineProps({
 	data: {
-		type: Object,
+		type: Array,
 		required: true
 	}
 })
+
+// const test = () => {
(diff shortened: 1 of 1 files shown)


LTDakin (approved):
LGTM

mgdaily (approved):
Nice small change!

PR #55, opened 2024-03-01
polling for operation completion

LTDakin:
Adds a watcher on props.operation to add polling intervals to each. Polls operations for completion and represents it visually with a loading bar under the button. Cleans up intervals after polling completes. Scoping this story down to just the polling logic, improvements on how it is visualized will be on a different PR.

Loading bar appears like in the top image and then is cleared when the object finished loading
Screenshot 2024-03-01 at 11 01 53 AM
This image is to highlight the polling process
Screenshot 2024-03-01 at 10 53 31 AM

### DIFF
--- a/src/components/DataSession/OperationPipeline.vue
+++ b/src/components/DataSession/OperationPipeline.vue
@@ -1,10 +1,13 @@
 <script setup>
-import { ref, defineEmits, defineProps} from 'vue'
+import { ref, defineEmits, defineProps, watch, onBeforeUnmount} from 'vue'
+import { useStore } from 'vuex'
+import { fetchApiCall, handleError } from '@/utils/api'
 import OperationWizard from './OperationWizard.vue'
 
+const store = useStore()
 const emit = defineEmits(['addOperation'])
 
-defineProps({
+const props = defineProps({
 	operations: {
 		type: Array,
 		required: true
@@ -17,6 +20,11 @@ defineProps({
 
 const selectedOperation = ref(-1)
 const showWizardDialog = ref(false)
+const operationPollingTimers = {}
+const operationPercentages = ref({})
+const POLL_WAIT_TIME = 1000
+const DEC_TO_PERCENT = 100
+const COMPLETE_PERCENT = 100
 
 function selectOperation(index) {
 	if (index == selectedOperation.value) {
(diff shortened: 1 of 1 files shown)


src/components/DataSession/OperationPipeline.vue
```diff
@@ -36,6 +44,40 @@ function operationBtnColor(index) {
 	}
 }
 
+async function pollOperationCompletion(operationID) {
+	const url = await store.state.datalabApiBaseUrl + 'datasessions/' + props.data.id + '/operations/' + operationID + '/'
+
+	const response = await fetchApiCall({ url: url, method: 'GET', failCallback: handleError })
+
+	if (response && response.percent_completion !== undefined && response.status != 'COMPLETED') {
```
capetillo, line 52: I think this can be the successCallback. e.g.:
```
returnOperationPercentage(data) {
if (response && response.percent_completion !== undefined && response.status != 'COMPLETED) {
etc...
}
```

src/components/DataSession/OperationPipeline.vue
```diff
@@ -36,6 +44,40 @@ function operationBtnColor(index) {
...
+	props.operations.forEach(operation => {
+		if (!operationPollingTimers[operation.id]) {
+			operationPollingTimers[operation.id] = setInterval( () => pollOperationCompletion(operation.id), POLL_WAIT_TIME)
+		}
+	})
+}, { immediate: true })
+
+onBeforeUnmount(() => {
+	// Clean up Polling Intervals
+	Object.keys(operationPollingTimers).forEach(operationID => {
```
capetillo, line 76: I think this needs an existence check for length

src/components/DataSession/OperationPipeline.vue
```diff
@@ -17,6 +20,11 @@ defineProps({
 
 const selectedOperation = ref(-1)
 const showWizardDialog = ref(false)
+const operationPollingTimers = {}
+const operationPercentages = ref({})
+const POLL_WAIT_TIME = 1000
```
capetillo, line 25: out of curiosity, is this in ms?

capetillo (commented):
a few comments and nice work overall

src/components/DataSession/OperationPipeline.vue
mgdaily: Will it crash on an empty object?

src/components/DataSession/OperationPipeline.vue
mgdaily: Agreed.

mgdaily (approved):
Nice! Looking good to me. Nice job re-scoping the story

src/components/DataSession/OperationPipeline.vue
capetillo: I tested on the console with an empty object like this
```
var x = {}
Object.keys(x).forEach(console.log("yes"))
```
and it returned 1 yes so I don't think it crashes but it returns erroneous data?

PR #56, opened 2024-03-01
Fix: Wizard styling buttons

capetillo:
### QUICK FIX: Wizard buttons position

Set the wizard buttons to have a position of fixed. For the screen recording, I changed `imagesPerRow` to be able to show that the buttons have a fixed position

**UPDATED AFTER LLOYD'S FEEDBACK**

https://github.com/LCOGT/datalab-ui/assets/54489472/555488b5-f4f8-4c1b-8228-ab1b01875aa2

Tablet views

https://github.com/LCOGT/datalab-ui/assets/54489472/385218fb-f63a-40c1-b63e-1286e332c78a

https://github.com/LCOGT/datalab-ui/assets/54489472/61cb59ba-b9e0-443b-92ea-e69e04d4dcc7

### DIFF
--- a/src/components/DataSession/OperationWizard.vue
+++ b/src/components/DataSession/OperationWizard.vue
@@ -275,9 +275,8 @@ const handleThumbnailClick = (item) => {
 }
 .operation-input {
   width: 10vw;
-  position: fixed;
   left: 3%;
-  top: 7%;
+  margin-bottom: 2%;
   background-color: var(--metal);
 }
 .buttons-container {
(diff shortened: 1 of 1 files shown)


src/components/DataSession/OperationWizard.vue
```diff
@@ -215,6 +215,7 @@ const handleThumbnailClick = (item) => {
 <style scoped>
 .wizard-background{
   background-color: var(--dark-blue);
+  height: 100vh;
```
LTDakin, line 218: The buttons are in the right place but from the video it still looks like the wizard component is overflowing and we have to scroll. For the tablet we don't want to have to scroll at all but want to see everything on one screen

LTDakin (changes requested):

src/components/DataSession/OperationWizard.vue
LTDakin: wasn't what I had in mind so actually we can go back to how it was before

LTDakin (approved):
can revert my request and it can be done another time, looks good though!

jnation3406 (commented):
I probably missed this in another PR, but the images there should have a header/label saying input_images or whatever the field key is in plain english, so they have an idea what they are selecting.

PR #57, opened 2024-03-01
Fix/reroute session postdeletion

LTDakin:
Deleting sessions will now send you to the first tab instead of just greeting you with a blank screen.

So a few changes
- added a check to loadSessions that helps when sessions are deleted. it checks whether our current tab is in the new data. if not then it sets tab to the first datasession
- just changed the v-model of v-tabs instead of using the router and onMounted to do so ( still can load a project from a url param, just doesn't reload the page when we change tabs ) 
- Selected tabs was not being used at all, guessing somewhere it got overwritten/replaced but wasn't removed
- line 26 is a bug I caught where you could put a param into the url of a data session that doesn't exist and then nothing would show up ( the later part of the video where you can't see my browser but I put in /999 and it handles it instead of being blank like before )

https://github.com/LCOGT/datalab-ui/assets/54085254/6126e5ef-87fc-44c3-a133-53b092009e8f

### DIFF
--- a/src/views/DataSessionsView.vue
+++ b/src/views/DataSessionsView.vue
@@ -14,7 +14,6 @@ const tab = ref()
 const deleteSessionId = ref(-1)
 const showDeleteDialog = ref(false)
 const dataSessionsUrl = store.state.datalabApiBaseUrl + 'datasessions/'
-const selectedTab = ref(-1)
 
 onBeforeMount(()=>{
 	if(!store.getters['userData/userIsAuthenticated']) router.push({ name: 'Registration' })
@@ -34,7 +33,7 @@ onMounted(async () => {
 		} else {
 			console.log('no data sessions available to display')
 		}
-		
+
 	}
 })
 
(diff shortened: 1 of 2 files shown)


src/views/DataSessionsView.vue
```diff
@@ -23,41 +22,38 @@ onBeforeMount(()=>{
...
 })
 
-function onTabChange(newSessionId) {
-	router.push({ name: 'DataSessionDetails', params: { sessionId: newSessionId }})
+function changeTab(newSessionId) {
+	tab.value = newSessionId
 }
 
-async function deleteSession(id) {
+function deleteDialog(id) {
```
capetillo, line 42: I think this should be named deleteSession. deleteDialog sounds like we delete the dialog

src/views/DataSessionsView.vue
```diff
@@ -23,41 +22,38 @@ onBeforeMount(()=>{
...
-}
-
-function selectTab(index) {
-	if (index == selectedTab.value) {
-		selectedTab.value = -1
-	}
-	else {
-		selectedTab.value = index
+	// if tab is not in new data default to displaying first tab
+	function updateData(data) {
```
capetillo, line 49: Is there a way in which this nested function can be its own separate successCallback?

src/views/DataSessionsView.vue
```diff
@@ -23,41 +22,38 @@ onBeforeMount(()=>{
...
 		} else {
 			console.log('no data sessions available to display')
 		}
-		
 	}
 })
 
-function onTabChange(newSessionId) {
-	router.push({ name: 'DataSessionDetails', params: { sessionId: newSessionId }})
+function changeTab(newSessionId) {
```
capetillo, line 38: Mild opinion, but I think this should keep the name onTabChange because it's describing what it's doing whereas changeTab sounds like that's what the function does

capetillo (commented):
some comments and a question. Great work!

capetillo (approved):
Nice! LGTM

PR #59, opened 2024-03-02
Styling/images per row

capetillo:
## STYLING: Updating the number of images per row based on screen size 

The title says it all. I added logic to check for the size of the window as well as an event listener and change the value of `imagesPerRow` based on that. The screen recording shows a big margin on top but that only comes up when the user has console open. I can write a separate story for that. 

https://github.com/LCOGT/datalab-ui/assets/54489472/190cc2c3-1e2e-4d31-97f6-f523848edda9


LTDakin (approved):
Nice functionality! Worth checking out flex rows because it might do something similar but I don't really know.

PR #60, opened 2024-03-02
styled delete session card

capetillo:
Styled the delete session card to look like the rest of the site : )
Screenshot 2024-03-01 at 4 44 58 PM


LTDakin (approved):
Looks good!

PR #64, opened 2024-03-05
Updated the ImageGrid and DataSession View to show output images

jnation3406:
CAROLINA WAKE UP and review this PR!!!

I'm not going to include a video, you can run it locally if you want. It adds the output images to the DataSession view, and it refactors the ImageGrid a bit to work generically with operation file inputs (not just those with key "input_files"), shows the operation that the image is from with operation Index in the upper right corner of the image (if it is output of an operation), and selection/image input is changed so that the correct image payload is sent to the server on selection. Also datasession view images can be filtered when clicking on an operation now.

There are so many changes because I had to run the auto-formatting on the files since it got mad that I dared to use spaces instead of tabs...

### DIFF
--- a/src/components/DataSession/DataSession.vue
+++ b/src/components/DataSession/DataSession.vue
@@ -1,5 +1,5 @@
 <script setup>
-import { defineEmits, defineProps } from 'vue'
+import { ref, defineEmits, defineProps, onMounted, watch } from 'vue'
 import OperationPipeline from './OperationPipeline.vue'
 import { fetchApiCall, handleError } from '../../utils/api'
 import { calculateColumnSpan } from '../../utils/common'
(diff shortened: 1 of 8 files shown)


src/components/DataSession/DataSession.vue
```diff
@@ -15,11 +15,78 @@ const props = defineProps({
...
+      filteredImages.value = images.value
+    }
+    else {
+      filteredImages.value = images.value.filter(image => image.operationIndex == operationIndex)
+    }
+  }
+}
 
 async function addOperation(operationDefinition) {
+<<<<<<< HEAD
```
capetillo: Delete please : )

src/components/DataSession/DataSession.vue
```diff
@@ -15,11 +15,78 @@ const props = defineProps({
...
+    }
+  }
+
+  // first operation doesn't load unless this is here
+  function postOperationSuccess() {
+    emit('reloadSession')
+  }
+
+  await fetchApiCall({ url: url, method: 'POST', body: operationDefinition, successCallback: postOperationSuccess, failCallback: handleError })
+=======
```
capetillo: Delete please : )

src/components/DataSession/DataSession.vue
```diff
@@ -31,36 +98,40 @@ async function addOperation(operationDefinition) {
   function postOperationSuccess(){
     emit('reloadSession')
   }
-	
+
   await fetchApiCall({url: url, method: 'POST', body: operationDefinition, successCallback: postOperationSuccess, failCallback: handleError})
+>>>>>>> fc904be8a779bf6f16316b0b95400427e2660111
```
capetillo: Delete again please

capetillo (commented):
Some left out comments and will need to talk to you to understand all of these changes thank u

src/components/DataSession/DataSession.vue
```diff
@@ -15,52 +15,107 @@ const props = defineProps({
...
 
+watch(() => props.data, () => {
+  props.data.operations.forEach((operation, index) => {
+    operationMap[operation.id] = index
+  })
+}, { immediate: true })
+
+
+onMounted(() => {
+  images.value = [...props.data.input_data]
```
LTDakin, line 98: setting these at lines 18 and 19 as arguments to the ref should be functionally the same

src/components/DataSession/DataSession.vue
```diff
@@ -15,52 +15,107 @@ const props = defineProps({
...
+  function postOperationSuccess() {
     emit('reloadSession')
   }
-	
-  await fetchApiCall({url: url, method: 'POST', body: operationDefinition, successCallback: postOperationSuccess, failCallback: handleError})
+
+  await fetchApiCall({ url: url, method: 'POST', body: operationDefinition, successCallback: postOperationSuccess, failCallback: handleError })
 }
 
+watch(() => props.data, () => {
```
LTDakin, line 90: watching objects can be expensive for vue, if you're only using the operations property here I think watching that property itself would work https://vuejs.org/guide/essentials/watchers.html#deep-watchers (second example ()=> state.someObject, state.someObject by itself won't work though it needs to be a getter which is what the anonymous function essentially is)

LTDakin (commented):
Just some small optimization things, otherwise nice job

LTDakin (approved):
lgtm

PR #66, opened 2024-03-06
Chore/unselect images

capetillo:
## CHORE: Unselect selected images
**NOTE** 
Eslint was making me change some files so I did that. The only file you should be looking at is `ProjectView`

**Background:**
When the user would select images and add them to a session, the selected images would remain selected. This has now been fixed.

**Implementation**
Added `onUnmounted` and committed `selectedImages` to have a value of an empty array. 

### Visuals
**Navigating from Project View to Data Session View and back**

https://github.com/LCOGT/datalab-ui/assets/54489472/7b98962e-e94a-49ed-9b52-c9437cb88b2d

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -8,29 +8,29 @@ const loadedConfig = computed(() => store.state.isConfigLoaded)
 
 // loading config  when app first mounts
 onMounted(async () => {
-	try {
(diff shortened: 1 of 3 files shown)


LTDakin (approved):
if I'm not missing anything the change is line 163 and 164. If so then looks good

PR #67, opened 2024-03-06
Style/image analysis

LTDakin:
Styled the Image Analysis component to match the rest of site as well as not overflow vertically on small screens 

ignore linting changes to RegistrationView.vue

Screenshot 2024-03-05 at 4 51 05 PM


capetillo (commented):
Being slightly nit-picky here: can you make the font: 
font-family: 'Open Sans', sans-serif;
and make the font of Details and the title of the image all capitals. You can do this by adding the property: 
text-transform: uppercase

mgdaily (approved):

PR #69, opened 2024-03-08
fix for local development content check

LTDakin:
I ran into this while developing with a local instance of the datalab backend.

After some testing there were two behaviors with DELETE. When the frontend hit the delete endpoint on the aws version of the datalab backend content-length does not exist and status-text is blank. However when the frontend hits the delete endpoint on a locally runnning datalab backend it does have content-length = 0. 

To solve this I added another check to the condition of blank content so that deleting will not error when working with a locally hosted version of datalab. 

The cause for the backend sending two different headers for the same operation is unclear however. Could be due to the aws endpoint being https and local being http.

Screenshot 2024-03-08 at 3 01 35 PM

### DIFF
--- a/src/utils/api.js
+++ b/src/utils/api.js
@@ -20,7 +20,7 @@ async function fetchApiCall({ url, method, body = null, header, successCallback 
   try {
     const response = await fetch(url, config)
     // ok response but empty content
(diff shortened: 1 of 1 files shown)


jnation3406 (approved):
looks fine to me

PR #70, opened 2024-03-11
Feature/filter by proposal

capetillo:
## FEATURE: Filter by proposal id 
**Background:**
In the future, the kids will probably have dozens (maybe hundreds?) of proposals/projects. I added logic to be able to filter by `proposal_id` so that the kids can easily search for their projects

**Implementation:**
First, I added a ref called `searchQuery` that checks for the user's input in the new `v-text-field`. 
Then, I defined a computed property called `filteredProjects` to filter projects based on the search query.
I first reduce to accumulate projects that match the `proposal_id` based on the `searchQuery`. If any projects match, add  them to the accumulator and then return the acc for the next iteration. 
Lastly, return the filtered list of projects.
If nothing is found, then just return all of the projects

### VISUALS
**Screen recording of successful filtering and everything still working as usual**

https://github.com/LCOGT/datalab-ui/assets/54489472/2a7ccfc1-f636-44da-9bc4-dc391c43fc24

### DIFF
--- a/src/components/Project/ProjectBar.vue
+++ b/src/components/Project/ProjectBar.vue
@@ -1,8 +1,8 @@
 <script setup>
-import { defineEmits, defineProps } from 'vue'
+import { defineEmits, defineProps, ref } from 'vue'
 import ProjectSelector from './ProjectSelector.vue'
 
-
+const searchQuery = ref('')
 const emit = defineEmits(['selectedProject'])
 
 defineProps({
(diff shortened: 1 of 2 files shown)


capetillo:
> Looks good. Some things that would be nice but not part of this PR - its a little confusing that no project on the left is selected initially even though one is shown on the right with images, and this becomes even more confusing when you can filter and have a project showing that no longer exists. Perhaps it would work to just include the name of the current filtered project somewhere to know what you are currently looking at on the right? Another thing is with the selection. It's nice that it maintains the selection state over switching between multiple projects, but then it becomes confusing what / if you have selections you are adding while switching projects. This could be solved by including some text next to the Add to Session button that says how many images you will be adding, or switching that buttons text to Add # images to Session. Then we would probably also want a button to clear the selection, in case they have added a ton of different images but decide against that... Just some ideas for future functionality.

I'll write a story!

PR #71, opened 2024-03-11
Chore/fixed buttons dspopup

capetillo:
## STYLING CHORE: Added fixed position to both buttons and input field in "add to data session" pop up
**Background:**
To improve QOL of users, I added a fixed position to the `cancel` and `create new session` buttons as well as the the input field to create a new data session. Additionally, I added scrolling styling to the list of data sessions.

**Implementation**
All CSS

### VISUALS
**Screen recording of scrolling through data sessions**

https://github.com/LCOGT/datalab-ui/assets/54489472/2c6f6451-7036-4094-9187-7bca62a273a9

### DIFF
--- a/src/components/Project/ProjectBar.vue
+++ b/src/components/Project/ProjectBar.vue
@@ -31,6 +31,7 @@ const selectProject = (projects) => {
   const proposalId = projects.map(p => p.proposal_id)
   emit('selectedProject', proposalId)
 }
+
 </script>
 
 <template>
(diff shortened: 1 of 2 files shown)


jnation3406 (changes requested):
I believe you've accidently inserted a bunch of whitespace at the ends of a bunch of lines (the green bars in the PR)? 

Otherwise looks fine with the caveat that I think we should switch the order of the buttons, i.e. cancel on the left, add on the right, to follow a left to right go back / go forward flow...

jnation3406 (approved):

PR #73, opened 2024-03-12
added unselect images logic

capetillo:
## FEATURE: Unselect images button
**Background:**
To facilitate the un-selection of all images, I added a button that resets `selectedImages` to `[]`.

**Implementation:**
Added quick logic that was already implement on `onUnmounted` but made it its own function (`unselectImages`). I'm struggling on whether this button should be conditionally rendered or if it should even be a button. But in the meantime, here's a functional button 

https://github.com/LCOGT/datalab-ui/assets/54489472/17abed0c-c9d4-4a97-917a-d52f632e3ba5

### DIFF
--- a/src/views/ProjectView.vue
+++ b/src/views/ProjectView.vue
@@ -73,6 +73,10 @@ const imageCounter = computed(() => {
   return store.state.selectedImages.length
 })
 
+const unselectImages = () => {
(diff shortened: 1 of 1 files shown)


jnation3406 (approved):
Looks good to me

PR #74, opened 2024-03-12
Feature/dsname character limit

capetillo:
## FEATURE: Data Session names now have a character minimum and max
**Background:**
Before, the user could create a data session and name it anything. It was pretty reminiscent of The Wild West. Now, there's a minimum and a maximum (pretty simple to change too). Styled it a bit so that it looks nice and aligned :) 

**Implementation:**
Added `validateSessionName` that checks for the length and name of the data session name. While the criteria for a correct data session name isn't met, it will display the errors. 

**IDEA:** add a counter?

### VISUALS
**Screen recording of name existing, being too short, existing, then fitting the criteria, then being too long**

https://github.com/LCOGT/datalab-ui/assets/54489472/b21bc80d-663d-4f66-b05e-4deef9294ed2

### DIFF
--- a/src/views/ProjectView.vue
+++ b/src/views/ProjectView.vue
@@ -13,7 +13,6 @@ const isPopupVisible = ref(false)
 const uniqueDataSessions = ref([])
 const newSessionName = ref('')
 const errorMessage = ref('')
-const nameLength = ref('')
 const isLoading = ref(true)
 const dataSessionsUrl = store.state.datalabApiBaseUrl + 'datasessions/'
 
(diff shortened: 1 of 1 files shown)


jnation3406 (changes requested):
It's a little confusing having the error stay there while you fix the problem. So only checking on button click leads to that confusion i think - a better way would be to have it check and display the error or not based on input to the field (typing anything). That way error would appear/disappear immediately based on whats in the field.

capetillo:
> It's a little confusing having the error stay there while you fix the problem. So only checking on button click leads to that confusion i think - a better way would be to have it check and display the error or not based on input to the field (typing anything). That way error would appear/disappear immediately based on whats in the field.

I have been led to more confusion

jnation3406 (approved):
Great!

PR #75, opened 2024-03-12
added logic to close popup and give newsessionname value ''

capetillo:
## CHORE: Close pop up and remove new session name on cancel 

**Background:**
More QOL chores. This one removes the session name that the user had written if they select cancel.

**Implementation:**
```
const closePopup = () => {
  isPopupVisible.value = false
  newSessionName.value = ''
  errorMessage.value = ''
}
```

that's all :) 

**EDIT: HERE'S THE CORRECT VIDEO**

https://github.com/LCOGT/datalab-ui/assets/54489472/26814270-e0ec-45b1-84cb-8254e70f9d97

### DIFF
--- a/src/views/ProjectView.vue
+++ b/src/views/ProjectView.vue
@@ -146,6 +146,11 @@ const validateSessionName = () => {
   }
 }
 
+const closePopup = () => {
+  isPopupVisible.value = false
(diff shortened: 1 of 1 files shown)


jnation3406 (approved):
Looks fine but your video does not show this feature

PR #79, opened 2024-03-18
Chore/add input text

capetillo:
## CHORE: Added `input images` text to Wizard

**Background**
We needed to add text to the wizard so that users know what they're doing or where they are

Screenshot 2024-03-18 at 2 33 32 PM

### DIFF
--- a/src/components/DataSession/OperationWizard.vue
+++ b/src/components/DataSession/OperationWizard.vue
@@ -272,10 +272,12 @@ function selectImage(inputKey, imageIndex) {
   font-size: 1rem;
 }
 
-.wizard-card {
-  width: 100%;
-  display: flex;
-  flex-direction: column-reverse;
+.operation-input {
+  width: 10vw;
+  margin-left: 2%;
+  background-color: var(--metal);
+  position: fixed;
+  top: 6%;
 }
 
 .images-container {
@@ -286,6 +288,7 @@ function selectImage(inputKey, imageIndex) {
   width: 100%;
   padding-left: 2rem;
   padding-right: 2rem;
+  margin-top: 6%;
 }
 .input-images {
   font-family: 'Open Sans', sans-serif;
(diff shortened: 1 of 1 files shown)


src/components/DataSession/OperationWizard.vue
```diff
@@ -158,25 +171,55 @@ function selectImage(inputKey, imageIndex) {
...
+            :label="inputDescription.name"
+            :type="inputDescription.type"
+            class="operation-input"
+          />
+          <div
+            v-else-if="inputDescription.type == 'file'"
+            class="images-container"
+          >
+            <div class="input-images">
+              Input Images
```
jnation3406, line 195: This label should be pulled from the inputDescription.name (we can have multiple file inputs and don't want them all just saying "Input Images". Also it would be nice to style the label exactly the same as the label on other inputs, i.e. "Duration" in your screenshot.

jnation3406 (changes requested):
You're IDE really likes to expand elements... it took me a while to spot the change.

src/components/DataSession/OperationWizard.vue
capetillo: my IDE is her own person and we respect her. Thanks for the feedback! done :)

src/components/DataSession/OperationWizard.vue
```diff
@@ -158,25 +171,58 @@ function selectImage(inputKey, imageIndex) {
...
+          />
+          <div
+            v-else-if="inputDescription.type == 'file'"
+            class="images-container"
+          >
+            <div
+              v-if="inputDescription.name == 'Input Files'"
+              class="input-images"
+            >
+              Input Files
```
jnation3406, line 198: No no I think you misunderstood. I want the text to be `{{ inputDescription.name }}` instead of "Input Files", that way it will show as "Input Files" in this case (where the name is Input Files), but for future operations, it will show whatever their input is called, without having to add a new case to the frontend. Does that make sense?

jnation3406 (changes requested):

src/components/DataSession/OperationWizard.vue
capetillo: yep yep, apologies

jnation3406 (approved):

PR #80, opened 2024-04-02
Feature/line analysis

capetillo:
## FEATURE: Plotting a line 

LONG PR

**Background:**
In order to get a line profile, users have to be able to draw a line on an image and have the x, y coordinates of the start and end points. This is that first step. 

**Implementation:**
After much sweat, blood, and tears (i.e. research) I found [Leaflet](https://leafletjs.com/reference.html#map-example) and a fantastic plugin for Leaflet, [Geoman](https://www.geoman.io/docs/). These two are compatible for tablet and work quite smoothly. I'm going to start from the top of `ImageViewer` and won't go into everything because the comments should explain what I'm doing. Also, I deleted `ZoomableImage` since Leaflet takes care of zooming in and out.

`loadImageOverlay`: dynamically loads the image into a Leaflet map as an overlay and adjusts the map to fit the image bounds. It also replaces any existing overlay with the displayed image in order to have the image's view centered and zoomed to display the entire image within its visible area. This is invoked `onMounted` and takes `props.imageSrc` as args so that it can adjust the size of the Leaflet map.

`displayLineProfile` and `getLineProfile`: both temporarily bogus. The latter is invoked after a line is drawn and takes the `startPoint` and `endPoint` as params. The success callback is the former and it only logs the data. 

The first step on `onMounted` is to create a Leaflet map (but referred to as image hereinafter) with initialized values. 
Then I created a custom control so that if the user is zoomed in and dragged in a corner, they can just click it and reset the view to center and the initial zoom. 

`image.on('zoomend'`: prevents image from being dragged if not zoomed and centers the image when user zooms out

`image.on('pm:drawstart'`: Finishes drawing a line after two points are selected and removes the last drawn line if another line is drawn

`image.on('pm:create'`: gets coordinates of the line and adjusts coordinate values based on zoom level. `getLineProfile` is invoked here with the start and end coordinates as args.

### VISUALS

**Screen recording of everything working accordingly**

https://github.com/LCOGT/datalab-ui/assets/54489472/96056721-cedc-40df-acfc-7c3db70f1fde

**Screenshot of console (read last log) with start and endpoints when a line is drawn at the very top left of the image**
Screenshot 2024-04-01 at 5 29 28 PM


src/components/Global/ImageGrid.vue
capetillo, line 1: For whatever reason, this is always marked as an error on my eslint so this is not really necessary but it wouldn't let my code run otherwise.

src/components/Project/ImageViewer.vue
```diff
@@ -0,0 +1,210 @@
...
+}
+
+const displayLineProfile = (data) =>{
+  if (data) {
+    console.log(data)
+  }
+}
+
+async function getLineProfile(startPoint, endPoint) {
+  // const url = 'http://localhost:8000/api/line-profile/'
```
mgdaily, line 59: Can we add a TODO just so it's clear this commented code is for calling the backend to get the line profile data?

mgdaily (approved):
Clean! Really excited about this new library

src/components/Project/ImageViewer.vue
capetillo: yes!

PR #82, opened 2024-04-06
Switch to Pinia cache, add in caching of thumbnails

jnation3406:
Attaching you all as you should all be aware of the vuex -> pinia change at least.

This PR removes vuex, adds pinia, and adds in caching of thumbnails into browser cache. It also uses a separate LRUCache to keep track of lifetimes of the cached image requests, with a lifetime set at 7 days, or thousands of images right now. If that lifetime is exceeded, the cached image should be automatically deleted and redownloaded/recached from the source.

Also I made a small change to set the active datasession tab, so that we only make those operation polling requests for operations on the active datasession.


babel.config.js
```diff
@@ -1,5 +1,8 @@
 module.exports = {
   presets: [
-    '@vue/cli-plugin-babel/preset'
+    '@vue/cli-plugin-babel/preset',
+  ],
+  plugins: [
+    '@babel/plugin-transform-private-methods'
```
jnation3406, line 6: This is needed for the LRUCache library to work

src/stores/thumbnails.js
```diff
@@ -0,0 +1,154 @@
...
+
+
+const cacheOptionsSmall = {
+  max: 5000,
+  dispose: (value, key, reason) => {
+    if (reason != 'set'){
+      deleteCachedImageFromBasename('small', key)
+    }
+  },
+  noUpdateTTL: true,
```
jnation3406, line 13: This lets us re-set the values of cache keys on reload without updating their TTL, which is necessary to do because generated URLs for the cached blobs are only valid for the lifetime of your browser window, i.e. they expire on reload.

src/stores/thumbnails.js
```diff
@@ -0,0 +1,144 @@
...
+        var largeThumbnailsCache2 = new LRUCache(cacheOptionsLarge)
+        largeThumbnailsCache2.load(tempValue.largeThumbnailsCache)
+        // Clear existing keys URL since they will be invalid on reload
+        for (const key of [...largeThumbnailsCache2.keys()]) {
+          const value = largeThumbnailsCache2.peek(key)
+          URL.revokeObjectURL(value)
+          largeThumbnailsCache2.set(key, '')
+        }
+        tempValue.largeThumbnailsCache = largeThumbnailsCache2
+        return (tempValue)
```
jnation3406, line 77: Special code is needed to serialize/deserialize the LRUCache because only simple data can be stored in browser localStorage by the pinia persistedstate plugin. We also clear out the old expired generated blob URLs on reload here, to prepare for regenerating them in the afterRestore callback

LTDakin (approved):
Wow great job and nice QOL improvments also. Scoped stores and the Image Loader util will be really helpful for the future.

src/stores/thumbnails.js
```diff
@@ -0,0 +1,143 @@
...
+    serializer: {
+      serialize: value => {
+        let serializedValue = {
+          archives: value.archives,
+          smallThumbnailsCache: value.smallThumbnailsCache.dump(),
+          largeThumbnailsCache: value.largeThumbnailsCache.dump()
+        }
+        return JSON.stringify(serializedValue)
+      },
+      deserialize: value => {
```
mgdaily, line 57: I think a simple comment here like:

```
// De-serialize image caches from persistent storage, invalidate object URLs, and refresh LRU caches on browser reload
```

would just help us when we're looking at this code in a years' time

src/stores/thumbnails.js
```diff
@@ -0,0 +1,143 @@
...
+    return {
+      archives: [],
+      smallThumbnailsCache: new LRUCache(cacheOptionsSmall),
+      largeThumbnailsCache: new LRUCache(cacheOptionsLarge),
+    }
+  },
+  persist: {
+    key: 'thumbnails',
+    serializer: {
+      serialize: value => {
```
mgdaily, line 49: A comment here saying simply

```
// Load LRU caches into persistent storage
```

would help

src/utils/imageLoader.js
```diff
@@ -0,0 +1,61 @@
...
+  })
+}
+
+async function getImageFromBasename(size, archive, url, imageBasename) {
+  // Clean up the -large or -small on the basename from ptr archive images
+  imageBasename = imageBasename.replace('-small', '').replace('-large', '')
+  const cacheName = size + 'Thumbnails'
+  return caches.open(cacheName).then((cache) => {
+    return cache.match(imageBasename, cacheOptions).then(function (response) {
+      return response || cacheImageFromBasename(size, archive, url, imageBasename)
```
mgdaily, line 22: I think a comment here saying something like

```
// if not found in cache, then fetch image information from archive and cache the result
```

would help

src/utils/imageLoader.js
```diff
@@ -0,0 +1,61 @@
...
+      })
+      return imageResponse.clone()
+    }
+    else {
+      console.error('Failed to get frame data for image ' + imageBasename)
+    }
+  }
+}
+
+export { getImageFromBasename, deleteCachedImageFromBasename, getAllCacheKeys }
```
mgdaily, line 61: Add a newline here

src/stores/settings.js
```diff
@@ -0,0 +1,76 @@
...
+      const imageUrl = option ? `${baseUrl}?${option}` : baseUrl
+	
+      const responseData = await fetchApiCall({ url: imageUrl, method: 'GET' })
+      if (responseData && responseData.results) {
+        // Preload all the small thumbnails into the cache. The large thumbnails will be loaded on demand
+        responseData.results.forEach((frame) => {
+          if (option === 'reduction_level=95'){
+            frame.cachedUrl = ref('')
+            thumbnailsStore.cacheImage('small', 'ptr', frame.url, frame.basename).then((cachedUrl) => {
+              frame.cachedUrl.value = cachedUrl
```
mgdaily, line 63: Maybe I'm misunderstanding but is this .then() callback needed? By this time we've cached the image, right? And we're done with the frame variable?

mgdaily (approved):
I think I've convinced myself I mostly understand what's going on...Just a few questions and suggestions for clarifying comments

src/stores/settings.js
LTDakin: I think `cacheImage()` is returning a new url that points to browser cache. So he's caching the image -> getting a url for the cache -> updating the response object that points to the cache with the right url. 

But I don't understand the need for `frame.cachedUrl = ref('')`
Can you do `frame.cachedUrl = cachedUrl` and delete line 61, making it into a reactive variable?

jnation3406: So the v-img components use the `cachedUrl` property as their 'src', so by initially setting it to an empty ref and then updating its value, this triggers the UI to automatically "show" the image once it is available. If you just set the cachedUrl without making it a ref then the image won't show up on initial load if its being cached/created then.

src/components/Global/ImageGrid.vue
```diff
@@ -71,8 +61,8 @@ watch(() => props.images, () => {
       :cols="columnSpan"
     >
       <v-img
-        v-if="image.basename in imageDetails"
-        :src="imageDetails[image.basename].url"
+        v-if="image.basename in imageDetails && imageDetails[image.basename]"
```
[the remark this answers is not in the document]
jnation3406: I'm going to leave it as-is... partly because lodash isn't a pulled into that component yet, and partly because I'm not sure if lodash isNil checks for empty string within a ref...

src/stores/thumbnails.js
```diff
@@ -0,0 +1,143 @@
...
+    setArchives(archives) {
+      this.archives = archives
+    },
+    async cacheImage(size, archive, url, basename, persist=true) {
+      basename = basename.replace('-small', '').replace('-large', '')
+      let cachedUrl = this.cachedUrl(size, basename)
+      if (cachedUrl === undefined || cachedUrl === ''){
+        return getImageFromBasename(size, archive, url, basename).then((response) => response.blob())
+          .then((responseBlob) => {
+            const responseUrl = URL.createObjectURL(responseBlob)
```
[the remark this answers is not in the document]
jnation3406: Yes, they generate a URL for a local resource so you can use it within an html component like ... But they are only guaranteed to live for the duration of your browser window, which is why the need to be regenerated on reload.

PR #86, opened 2024-04-26
Fix adding operations not immediately polling for completion, and add…

jnation3406:
… in showing datalab generated images

blurry median image shown


LTDakin (approved):
lgtm

PR #87, opened 2024-05-02
Line Profile and Analysis Actions

LTDakin:
Updating the UI code to be able to work with the new endpoint for drawing line profiles in the Image Analysis component

Added a general function request analysis that takes some input and action name and calls the backend for it. This way we can easily add new actions in the future and call the same function.

Let the parent component handle the api call so that the line plot and image viewer can be more granular in terms of function. The image viewer emits an AnalysisAction event with the action name line-profile in this case. The ImageAnalyzer parent will then call the backend /analysis endpoint with that payload. When it receives a response it will handle it in a switch case based on which action it got back. In this case since its a lineProfile we update the ref variables and the linePlot has a watcher on those variables that will redraw its graph.

Analysis components were also grouped to be in their own folder inside projects for organization.

Screenshot 2024-05-02 at 10 18 18 AM


src/components/Project/ImageAnalysis/ImageViewer.vue
```diff
@@ -164,6 +130,17 @@ onMounted(() => {
...
+    lastDrawnLine = e.layer
+    e.layer.on('pm:edit', handleEdit)
+    findPixelCoords(e.layer.getLatLngs())
+  })
+}
+
+// Callback function for when a line is edited and checks if it's within bounds
+function handleEdit(event) {
   // Checks if the point is within the bounds of the image
   function isWithinBounds(point) {
```
capetillo, line 145: this being inside the `handleEdit` does it get triggered when the user first draws a line? In other words, if the user draws a line and extends it outside of the image, does the endpoint stay within bounds?

src/components/Project/ImageAnalysis/ImageViewer.vue
```diff
@@ -176,78 +153,73 @@ onMounted(() => {
...
-  }
-  // Get coordinates of the line
-  image.on('pm:create', (e) => {
-    // Save last drawn line
-    lastDrawnLine = e.layer
-    e.layer.on('pm:edit', handleEdit)
-    getLineLength(e.layer.getLatLngs())
-  })
+onMounted(() => {
+  loadImageOverlay()
```
capetillo, line 221: nice refactoring! this definitely makes it more readable : )

src/components/Project/ImageAnalysis/LinePlot.vue
```diff
@@ -1,13 +1,10 @@
...
-import { useSettingsStore } from '@/stores/settings'
+import { onMounted, ref, watch } from 'vue'
 import * as d3 from 'd3'
 
 const svg = ref(null)
-const store = useSettingsStore()
 
-const randomNumbers = computed(() => store.randomNumbers)
-const lineLength = computed(() => store.lineLength)
+const props = defineProps(['yAxisData', 'xAxisSize'])
```
capetillo, line 7: nit-picking here, I think these could be named more clearly. Maybe something like `luminosity` and `arcsecs` or if you want to include the y and x values to make it knowns what's what, `yAxisLuminosity` and `xAxisArcsecs`

capetillo (commented):
Great work! I left a few comments and one question :-)

src/components/Project/ImageAnalysis/ImageViewer.vue
mgdaily, line 145: Is there a reason we need this function within another function?

LTDakin: I moved it outside at Matt's request and tested it again and looks like it is working great, The point is staying inside the bounds of the image even when the user tries to draw it outside.

LTDakin: moved it

src/components/Project/ImageAnalysis/LinePlot.vue
LTDakin: agreed, since this plot is tuned for the line analysis I went with your second suggestion. In the future if we have another graph though maybe we can add two more props to this component yAxisName and xAxisName so we can reuse it.

capetillo (approved):
nice, lgtm! thank you for the changes

PR #88, opened 2024-05-03
Setup datalab ui to connect to LCO obs portal and archive and use thu…

jnation3406:
…mbnail service

I also separated out the initial configuration into its own pinia store for clarity. I changed the behavior of loading data so that it loads for a specific proposal/project it as you click on a that proposal/project. It will also regrab the frames when the time window is changed. This should continue to work with the PTR test archive if you set archiveType to "ptr" in the config.json and point back to the PTR services.

The referenced datalab.lco.global isn't running yet, but I'm working on getting that up in the LCO cluster so you could use that to test against. For now you can run the datalab backend locally and point to that, as long as you make an account with an api token in your local datalab that matches your observation portal username/api token. In production, it will automatically have all the observation portal accounts.


src/stores/settings.js
```diff
@@ -48,28 +33,25 @@ export const useSettingsStore = defineStore('settings', {
...
       const thumbnailsStore = useThumbnailsStore()
-      const baseUrl = this.datalabArchiveApiUrl + 'frames/'
+      const configurationStore = useConfigurationStore()
+      const baseUrl = configurationStore.datalabArchiveApiUrl + 'frames/'
+      const timeStr = `start=${this.startDate.toISOString()}&end=${this.endDate.toISOString()}`
+      option = option ? `${option}&${timeStr}` : timeStr
+      option += '&proposal_id=' + proposal
       const imageUrl = option ? `${baseUrl}?${option}` : baseUrl
-	
+      console.log('Getting frames from: ' + imageUrl)
```
capetillo, line 44: Was this used for testing?

capetillo (commented):
Great work! It looks good to me, just one console.log is there that I don't know if you intended to leave or forgot about it : )

LTDakin (approved):
other than that console log Carolina pointed out, nice work

PR #89, opened 2024-05-22
User Logout and Browser saved settings

LTDakin:
added a new component called usermenu that displays who is currently logged in, a logout button, the colorblind -> high contrast, and grid/carousel settings. 

These settings are moved to the userData store since its persisted and the settings store seems to be scoped for project settings (possible rename in the future needed). 

Used vuetify to rewrite the navbar since I can't float an item to the right to save my life in css and also to keep it consistent with using vuetify components where we can. 

Also for the registration.vue I just added the var metal color to the card so it matches with the rest of the site, but there was eslint formatting that spaced out all the attributes. 

Screenshot 2024-05-22 at 2 03 40 PM

Screenshot 2024-05-22 at 2 04 14 PM

https://github.com/LCOGT/datalab-ui/assets/54085254/c4fce490-2391-4823-8ded-409d161804bd


LTDakin:
@jnation3406 is this good for now or do you think I should still make a complete profile page for settings. I was torn since right now we only have two settings.

jnation3406 (approved):
Looks great to me, and so fast!

capetillo (approved):
Looks so nice! Great work, Lloyd

PR #90, opened 2024-06-05
source-catalog analysis action 

LTDakin:
This PR adds the source catalog analysis functionality to datalab ui. 

Now on the leaflet setup we call a function that requests a source catalog from the backend, the resulting response is passed back to the imageViewer in a prop called catalog which is being watched for updates. When it updates the code will create a new LayerGroup containing circles that mark the sources on the image.

We also have a new LayerControl which lets us dynamically add layerGroups to it to let us toggle them on and off.

Removed the code that was changing LatLngs to pixel coords and the reset on zoom since to make the line profile and source catalog display accurately for the image. Even though it sounds like it should be the image pixel coords I guess the LatLng from leaflet is shared with the fits coords, so thats good to remember for the future. 

https://github.com/LCOGT/datalab-ui/assets/54085254/c76042db-aeaf-4ecf-b5d2-0250eeeab981

Screenshot 2024-06-05 at 1 03 03 PM


src/components/Project/ImageAnalysis/ImageViewer.vue
```diff
@@ -23,6 +28,7 @@ let imageBounds
 let initialCenter = [0, 0]
 let initialZoom = 1
 let lastDrawnLine = null
+let LayerControl = null
```
capetillo, line 31: minor thing, I would name this `layerControl` to keep consistency

src/components/Project/ImageAnalysis/ImageViewer.vue
```diff
@@ -39,16 +45,23 @@ const loadImageOverlay = () => {
     minZoom: 0,
     dragging: false,
   })
-  
+
+  // Can add new layers to this control and they will be able to be turned on and off in the top right corner of the map
+  LayerControl = L.control.layers().addTo(image)
```
capetillo, line 50: and I would change it here

src/components/Project/ImageAnalysis/ImageViewer.vue
```diff
@@ -179,33 +188,15 @@ function handleEdit(event) {
   // Re-attach the edit handling logic
   newLine.on('pm:edit', handleEdit)
 
-  findPixelCoords(event.layer.getLatLngs())
+  requestLineProfile(event.layer.getLatLngs())
 }
 
-// Refactored code to calculate line length in pixels
-function findPixelCoords(latLngs) {
+function requestLineProfile(latLngs) {
```
capetillo, line 194: how do `startCoordinates` and `endCoordinates` know what the start and end are if the user has zoomed in on the image? does `getLatLngs` take care of it?

src/components/Project/ImageAnalysis/ImageViewer.vue
```diff
@@ -217,6 +208,37 @@ function findPixelCoords(latLngs) {
   emit('analysisAction', 'line-profile', lineProfileInput)
 }
 
+function createCatalogLayer(){
+  let catalogMarkers = []
+
+  props.catalog.forEach((source) => {
```
capetillo, line 214: I think this needs an existence check because `catalog` may be empty

capetillo (commented):
Nice work! Left a few comments and a couple of questions : )

src/components/Project/ImageAnalysis/ImageViewer.vue
LTDakin: Yeah! Did some testing to confirm but it seems like leaflet doesn't care about what zoom level you're at and will read the same lat lngs when you click on the same point on the image

src/components/Project/ImageAnalysis/ImageViewer.vue
LTDakin: oopsies good catch

LTDakin:
Added a screenshot of a zoomed in look at the overlay, also did a line profile zoomed in to demonstrate that it still has the right lat lngs @capetillo

src/components/Project/ImageAnalysis/ImageViewer.vue
capetillo: oh no way! that's awesome

capetillo (approved):
Awesome! Looks good to me

mgdaily (approved):

jnation3406 (approved):

PR #92, opened 2024-06-12
CSS, Behavior, Bug fixes

LTDakin:
Polishing up some parts of the datalab to make the site feel nicer to navigate and use

App.vue
- no longer shows the navbar when a user is not logged in

ProjectBar.vue
- shows the title above the time filter for easier viewing, now the flow is Title > time filter > search filter > results which is more intuitive

index, settings, datasessionView
- now tracks the last used data session, so the user will always see the data session they most recently worked on
- removed route param which fixes a ui bug with the navbar tabs

ProjectView
- fixed overflow on the page
- fixed buttons resizing differently
- made buttons more concise and added icons
- removed redundant css

Before

After

Before

After

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -36,8 +36,10 @@ watch(() => userDataStore.isColorblindMode, (newVal) => {
 </script>
 
 <template>
-  <nav-bar />
   <template v-if="loadedConfig">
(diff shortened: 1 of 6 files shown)


capetillo (approved):
Nicely done!

PR #94, opened 2024-06-26
More Readable Status handling

LTDakin:
This was some changes that came from me debugging and fixing the operation output issue, makes the handling of operation status to be more readable for the future, also fixes some ui placement with the operation buttons

### DIFF
--- a/src/components/DataSession/OperationPipeline.vue
+++ b/src/components/DataSession/OperationPipeline.vue
@@ -58,11 +58,14 @@ async function pollOperationCompletion(operationID) {
 
   const updateOperationStatus = (response) => {
     if(response){
-      switch(response.status){
+      const operationStatus = response.status
+      const percentCompletion = response.percent_completion
+
+      switch(operationStatus){
       case 'PENDING':
         break
       case 'IN_PROGRESS':
-        operationPercentages.value[operationID] = response.percent_completion * DEC_TO_PERCENT
+        operationPercentages.value[operationID] = percentCompletion * DEC_TO_PERCENT
         break
       case 'COMPLETED':
         operationPercentages.value[operationID] = COMPLETE_PERCENT
@@ -73,7 +76,7 @@ async function pollOperationCompletion(operationID) {
         }
         break
       default:
-        console.error('Unknown Operation Status:', response.status)
+        console.error('Unknown Operation Status:', operationStatus)
       }
     }
     else{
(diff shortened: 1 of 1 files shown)


src/components/DataSession/OperationPipeline.vue
```diff
@@ -56,20 +56,32 @@ function operationBtnColor(index) {
...
-  const updateOperationPercentages = (response) => {
-    if (response && response.percent_completion !== undefined && response.status != 'COMPLETED') {
-      operationPercentages.value[operationID] = response.percent_completion * DEC_TO_PERCENT
-    } else {
-      operationPercentages.value[operationID] = COMPLETE_PERCENT
-      emit('operationCompleted', response)
-      if (operationID in operationPollingTimers) {
-        clearInterval(operationPollingTimers[operationID])
-        setTimeout(clearPollingData(operationID), POLL_WAIT_TIME)
+  const updateOperationStatus = (response) => {
```
capetillo, line 59: Is it possible to pass the operation id and the response so that it doesn't have to be nested?

src/components/DataSession/OperationPipeline.vue
```diff
@@ -56,20 +56,32 @@ function operationBtnColor(index) {
...
+        console.error('Unknown Operation Status:', response.status)
       }
     }
+    else{
+      console.error('No response on status for operation:', operationID)
+    }
   }
 
-  await fetchApiCall({ url: url, method: 'GET', successCallback: updateOperationPercentages, failCallback: handleError })
+  await fetchApiCall({ url: url, method: 'GET', successCallback: updateOperationStatus, failCallback: handleError })
```
capetillo, line 84: So maybe here you can do updateOperationStatus(operationID, response). Something along those lines

capetillo (commented):
Just a comment about nesting, other than that looks good!

src/components/DataSession/OperationPipeline.vue
LTDakin: response is good anywhere since the callback will pass it to whatever function is there. 

Looking at operationID though it would be really tricky since its already passed as an argument to the parent function PollOperationCompletion, so to un-nest this function would mean adding a ref variable to track what pollOperationCompletion id is called.

[Good explanation for why certain functions make sense to nest ](https://stackoverflow.com/a/44482736/9627391) particularly the points 1, 2, and 5 for this case

capetillo (approved):
elle gee tee em

src/components/DataSession/OperationPipeline.vue
LTDakin: my mind completely missed this option...

PR #95, opened 2024-07-09
Global Alerts, Handling of failed operations, Wizard Styling changes

LTDakin:
Added a global alert component that shows a small toast at the bottom of the page. Added a store called `alerts` the alert toast listens to so if any component wants to be able to create alerts they just need to import the store and call it's `setAlert` function.

OperationPipeline polling status now has a check for `FAILED` operations. When an operation fails a couple things happen:
1. An Alert toast is shown with the error
2. We emit a `deleteSession` event to the parent `DataSessions`
3. The parent sends a delete api call to the backend to remove the failed operation and then reloads to update

Other Small things
- Noticed a lot of code reuse when clearing the polling data so bundled it all up into one function called `clearPolling(operationID)`
- Changed the way that the selected button is highlighted to use a vue3 class binding to remove the need for an extra function at the top of the file
- Moved the operation wizard from inside the operation pipeline as it didn't have any need to be a level down and was just passing the images prop through the pipeline. Now instead of `DataSessions` -> `OperationPipeline` -> `OperationWizard` it is `OperationWizard`  <- `DataSessions` -> `OperationPipeline` for better hierarchy. 
- Had to delete some of the css in the wizard since it was overflowing and the input boxes were unable to be seen when there was more than one. 

https://github.com/LCOGT/datalab-ui/assets/54085254/a826c846-ff99-4fe1-9a22-1495f95c5bcb

### DIFF
--- a/src/stores/alerts.js
+++ b/src/stores/alerts.js
@@ -1,0 +1,25 @@
+import { defineStore } from 'pinia'
+
+/**
+ * This store is used to manage the alert that is displayed at the top of the page.
+ * The alert can be of type 'success', 'error', 'warning', or 'info'.
+ * To use this import the store and call the setAlert action with the type and text of the alert.
+ */
+export const useAlertsStore = defineStore('alerts', {
+  state() {
+    return{
+      // The type of alert to display ('success', 'error', 'warning', 'info')
+      alertType: 'warning',
+      // The text to display in the alert
+      alertText: 'default alert text',
+    }
+  },
+  actions: {
+    setAlert(type, text, prefix = '') {
+      console.log('prefix', prefix)
+      console.log(prefix + text)
+      this.alertType = type
+      this.alertText = prefix ? prefix + ' ' + text : text
+    }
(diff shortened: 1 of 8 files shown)


src/stores/alerts.js
```diff
@@ -0,0 +1,25 @@
...
+    return{
+      // The type of alert to display ('success', 'error', 'warning', 'info')
+      alertType: 'warning',
+      // The text to display in the alert
+      alertText: 'default alert text',
+    }
+  },
+  actions: {
+    setAlert(type, text, prefix = '') {
+      console.log('prefix', prefix)
```
capetillo, line 19: I think you forgot to delete these two console logs !

capetillo (approved):
After deleting the console logs, LGTM! Nice work Lloyd 🥳

src/stores/alerts.js
LTDakin: those darn console logs, thank you!

PR #96, opened 2024-07-15
DataSession code is incorrect as it overides backend, operations wizard has nonused var

LTDakin:
Some code cleaning as I'm researching stories

- displayImages not used anywhere
- moved the ref to be at the top with the rest of the refs
- the if check is overwriting the source attribute of images, maybe it was for when the site was first starting up but we need source to be datalab as well for future stories

Tested to make sure everything still works

### DIFF
--- a/src/components/DataSession/DataSession.vue
+++ b/src/components/DataSession/DataSession.vue
@@ -85,11 +85,6 @@ function deleteOperation(operationID) {
 
 async function addOperation(operationDefinition) {
   const url = dataSessionsUrl + props.data.id + '/operations/'
-  if ('input_files' in operationDefinition.input_data) {
-    for (const image of operationDefinition.input_data.input_files) {
-      image.source = 'archive'
-    }
-  }
 
   // first operation doesn't load unless this is here
   function postOperationSuccess() {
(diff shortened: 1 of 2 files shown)


src/components/DataSession/DataSession.vue
```diff
@@ -85,11 +85,6 @@ function deleteOperation(operationID) {
 
 async function addOperation(operationDefinition) {
   const url = dataSessionsUrl + props.data.id + '/operations/'
-  if ('input_files' in operationDefinition.input_data) {
-    for (const image of operationDefinition.input_data.input_files) {
-      image.source = 'archive'
```
capetillo, line 90: Can't remember so: does `image.source` have an assigned value prior to this?

LTDakin: yeah its gets that property when its sent back from the archive or datalab, however this code was overwriting the image.source = 'datalab' with 'archive' which was wrong

capetillo (commented):
Left a question : ) so if `image.source` comes with a value, then all is good. But if it doesn't, there is an existence check below that could cause errors!

src/components/DataSession/DataSession.vue
capetillo: gotcha! then LGTM 🎉

LTDakin: we may need to still add a method for patching in missing sources but this code was arbitrarily replacing all image sources with 'archive' which isn't what we want to happen

capetillo (approved):
🤠

PR #97, opened 2024-07-30
Feature/rgb composite

LTDakin:
Lots of changes to support the RGB operation. Adds the `filter` component that lets users know what filters an image has. 
Implement the filtering for inputs to set their max and min selected images as well as what filters are required for inputs. 
The general idea is that users will be able to know what filters their images have and easily be able to create RGB operations while being constrained to entering only valid input. 

Other changes 
- Added `alerts` to operation wizard so when max and min inputs are failed the user knows
- `imagesWithFilter` function helps display only R,G,B images if an input requests for it
- add check for max images in an input
- `image-grid` prop `selected-images` is now tracked by the parent and passed to `image-grid` this is so the parent can have better control over what images show up as selected since they are the one implementing logic to decide what to do with the selected images
- `alert-toast` `z-index` set to `9999` so it can show over the `v-dialog` which has a `z-index: 2400`
- new `FilterBadge` component that decides what color to be depending on the filter name passed to it. When we add more supported colored filters in the future this should be updated
- `FilterBadge` is added to `imageCarousel` ,`imageList`, `imageAnalysis`, and `imageGrid` as these are all places users would be interested to know the filter type of the image
- fix download button not being the same color as the rest of the analysis page
- list view now fits screen and scrolls for a much nicer behavior
- add `filter` to image objects

Screenshot 2024-07-30 at 12 09 12 PM
Screenshot 2024-07-30 at 12 09 00 PM
Screenshot 2024-07-30 at 12 08 52 PM
Screenshot 2024-07-30 at 12 08 32 PM
Screenshot 2024-07-30 at 12 08 18 PM

### DIFF
--- a/src/components/DataSession/OperationWizard.vue
+++ b/src/components/DataSession/OperationWizard.vue
@@ -93,6 +93,13 @@ const wizardTitle = computed(() => {
 
 function goForward() {
   if (page.value == 'select') {
+    // if there are no images for a filter required by the operation, do not proceed
+    for (const inputKey in selectedOperationInputs.value) {
+      const inputField = selectedOperationInputs.value[inputKey]
+      if (inputField.type == 'file' && imagesWithFilter(inputField.filter).length == 0){
+        return
+      }
+    }
     page.value = 'configure'
   }
   else {
--- a/src/components/DataSession/OperationWizard.vue
+++ b/src/components/DataSession/OperationWizard.vue
@@ -91,14 +91,6 @@ const wizardTitle = computed(() => {
   }
 })
 
-function imagesWithFilter(filters) {
-  const imagesFiltered = props.images.filter((image) => filters.includes(image.filter))
-  if (imagesFiltered.length == 0) {
-    alert.setAlert('warning', 'Operation requires images with filter ' + filters.join(', '))
-  }
-  return imagesFiltered
-}
-
 function goForward() {
   if (page.value == 'select') {
     page.value = 'configure'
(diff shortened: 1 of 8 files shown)


src/components/DataSession/OperationWizard.vue
```diff
@@ -101,23 +103,42 @@ function goForward() {
...
       })
-      operationDefinition.input_data[inputKey] = selected
+      operationDefinition.input_data[inputKey] = input
     }
     emit('addOperation', operationDefinition)
     emit('closeWizard')
   }
 }
 
+function imagesWithFilter(filters) {
```
capetillo, line 118: this is just a suggestion for a better UX (imo and doesn't have to be implemented now but to consider in the future): instead of getting a warning, could the user just not select that operation if none of the images qualify for the rgb filter?

src/components/Global/AlertToast.vue
```diff
@@ -23,7 +23,7 @@ watch(() => alertStore.alertText, () => {showAlert.value = true})
     position: absolute;
     bottom: 0;
     left: 0;
-    z-index: 1000;
+    z-index: 9999;
```
capetillo, line 26: here's a hot tip: avoid not keeping track of your z-indexes. Only use 9999 for something you absolutely know that has to go on top of everything no matter what. Otherwise, I would suggest starting from z-index: 2 and going up from there. It's easier to do it now than later

src/components/Global/ImageGrid.vue
```diff
@@ -70,9 +62,12 @@ watch(() => props.images, () => {
         cover
         :class="{ 'selected-image': isSelected(index) }"
         aspect-ratio="1"
-        class="image-grid"
-        @click="onImageClick(index)"
+        @click="emit('selectImage', index)"
       >
+        <filter-badge
+          v-if="image.filter"
+          :filter="image.filter"
```
capetillo, line 69: I'm not sure if this is where the text in the upper left corner of the image is generated, but another suggestion for better UX, _personally_ I would prefer to read "red", "green", or "blue" in tiny letters OR nothing at all (though that's not color blind friendly) than just the letters of the filters. I'm not sure if by then, they'll know what the filters are and if it's relevant to just see the letters

capetillo (commented):
Left a few comments, suggestions, food for thought. Nice job getting this to work!

src/components/Global/AlertToast.vue
LTDakin: ahhh thats a good practice, I can change this one to 2401, the v-overlay component which was blocking it sits at 2400

src/components/DataSession/OperationWizard.vue
LTDakin: done!

src/components/DataSession/OperationWizard.vue
```diff
@@ -91,6 +93,13 @@ const wizardTitle = computed(() => {
 
 function goForward() {
   if (page.value == 'select') {
+    // if there are no images for a filter required by the operation, do not proceed
+    for (const inputKey in selectedOperationInputs.value) {
```
jnation3406, line 97: This will stop you from advancing with that operation, but really want you want is to grey out the "configure" button if that operation is selected so the user knows its not available. Or even better grey the button out and have some text in the description showing the unmet requirement. Perhaps this is a new story of functionality to add to the wizard of showing the requirements/preconditions that must be met for an operation to be selected.

jnation3406 (approved):
Looks good to me

PR #98, opened 2024-08-08
Add support for thumbnails in the archive

jnation3406:
Add a `smallThumbUrl` and `largeThumbUrl` to the archive image results - either empty if there are no thumbnails in the response, or the thumb url if it exists. Then use that throughout when caching them. The caching code already pulls either directly from a URL or from the thumbservice so if a URL is provided it will use that.

To actually test this, you have to set the reduction level to 0 for archive images you retrieve since we only have the thumbnails for raw images that are EXPOSE type...

### DIFF
--- a/src/components/Global/ImageGrid.vue
+++ b/src/components/Global/ImageGrid.vue
@@ -37,7 +37,7 @@ watch(() => props.images, () => {
   props.images.forEach(image => {
     if (image.basename && !(image.basename in imageDetails.value)) {
       imageDetails.value[image.basename] = ref('')
-      const url = image.thumbnail_url || ''
+      const url = image.thumbnail_url || image.smallThumbUrl || ''
       thumbnailsStore.cacheImage('small', configurationStore.archiveType, url, image.basename).then((cachedUrl) => {
         imageDetails.value[image.basename] = cachedUrl
       })
--- a/src/stores/settings.js
+++ b/src/stores/settings.js
@@ -35,13 +35,24 @@ export const useSettingsStore = defineStore('settings', {
       const timeStr = `start=${this.startDate.toISOString()}&end=${this.endDate.toISOString()}`
       option = option ? `${option}&${timeStr}` : timeStr
       option += '&proposal_id=' + proposal
+      option += '&include_thumbnails=true'
       const imageUrl = option ? `${baseUrl}?${option}` : baseUrl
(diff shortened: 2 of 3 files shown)


src/components/Global/ImageGrid.vue
```diff
@@ -37,7 +37,7 @@ watch(() => props.images, () => {
   props.images.forEach(image => {
     if (image.basename && !(image.basename in imageDetails.value)) {
       imageDetails.value[image.basename] = ref('')
-      const url = image.thumbnail_url || ''
+      const url = image.thumbnail_url || image.smallThumbUrl || ''
```
jnation3406, line 40: the first thumbnail_url is from images generated in the server

LTDakin: so it tries to first get images from the server and if that doesn't work where are .smallThumbUrl images coming from? the thumbnail service?

src/stores/settings.js
```diff
@@ -35,13 +35,24 @@ export const useSettingsStore = defineStore('settings', {
       const timeStr = `start=${this.startDate.toISOString()}&end=${this.endDate.toISOString()}`
       option = option ? `${option}&${timeStr}` : timeStr
       option += '&proposal_id=' + proposal
+      option += '&include_thumbnails=true'
       const imageUrl = option ? `${baseUrl}?${option}` : baseUrl
       const responseData = await fetchApiCall({ url: imageUrl, method: 'GET' })
       if (responseData && responseData.results) {
         // Preload all the small thumbnails into the cache. The large thumbnails will be loaded on demand
```
LTDakin, line 42: does this comment need changing since it seems to be pre-loading the large thumbnails as well

LTDakin (approved):
some questions

src/stores/settings.js
jnation3406: no its not preloading large thumbnails, its just storing the large and small thumbnail urls into the object if they exist in the response. But its only preloading the small ones into the cache.

src/components/Global/ImageGrid.vue
jnation3406: so smallThumbUrl is what is set in the settings.js `loadAndCacheImages` function now, if it is set in the archive response. Its a direct link to download the small thumbnail from S3. If it doesn't exist we just set empty string, which triggers us to use the thumbnail service as a fallback

PR #99, opened 2024-08-14
Feature/tif downloads

LTDakin:
New download menu to pick between downloading a FITs, Jpg, or TIF file.

To facilitate this there is now a `ImageDownloadMenu` component that emits a `get-tif` analysis action on mount. This is because images don't come with a tif file so we have to request the backend to generate one for us. 

Video showcasing the download menu

https://github.com/user-attachments/assets/5f91504c-c387-4db2-8cd3-264ae59219ea

Message when a file is missing
Screenshot 2024-08-14 at 12 53 58 PM


capetillo (approved):
This looks great! Nice work

src/components/Project/ImageAnalysis/ImageDownloadMenu.vue
```diff
@@ -0,0 +1,64 @@
...
+    'basename': props.image.basename
+  }
+  emit('analysisAction', 'get-tif', getTifInput)
+})
+
+</script>
+<template>
+  <v-speed-dial
+    location="left center"
+    transition="fade-transition"
```
jnation3406, line 35: I like the fade transition, but I wonder if it might make more sense having the menu go down from the button rather than to the left of the button?

src/components/Project/ImageAnalysis/ImageDownloadMenu.vue
```diff
@@ -0,0 +1,64 @@
...
+  const a = document.createElement('a')
+  a.href = link
+  a.download = filename
+  document.body.appendChild(a)
+  a.click()
+  document.body.removeChild(a)
+}
+
+onMounted(() => {
+  const getTifInput = {
```
jnation3406, line 25: It's not clear to me from the code but does this only call get tiff output once the user clicks the button? or does it call it when they enter the analysis view for an image? If its the former that is fine, but if its the later I don't think we want to generate the tiff everytime they just view the image in the analysis tab, we only want to have the backend generate the tiff on demand. I think the correct way would be to have the frontend call the backend to generate and return the tiff on click of the download tiff button.

src/components/Project/ImageAnalysis/ImageDownloadMenu.vue
```diff
@@ -0,0 +1,64 @@
...
+    />
+    <v-btn
+      key="2"
+      text=".TIF"
+      @click="downloadLink(props.tifUrl, props.image.basename, 'TIF')"
+    />
+    <v-btn
+      key="3"
+      text=".JPG"
+      @click="downloadLink(props.image.largeCachedUrl, props.image.basename, 'JPG')"
```
jnation3406, line 56: I think we might want to get the jpeg from the server as well, same way as the tiff... The largest jpeg we generate for the view is like 900px, but the full image might have a much higher resolution and I think downloading the jpeg you would want the full resolution. But again get it on demand rather than pre-generate it for every image.

jnation3406 (changes requested):
Looks good, the only thing I'm not sure about is if we are generating the tiffs on demand or not.

src/components/Project/ImageAnalysis/ImageDownloadMenu.vue
LTDakin: I chose the left position to keep the input from overlapping info below it and avoid finger blockage on tablets. When tapped, options are easily visible without needing to lift your finger. While placing it below is more common, I think the advantages of the left position are worth it

src/components/Project/ImageAnalysis/ImageDownloadMenu.vue
LTDakin: I think we're going to replace the carousel with the grid in which case theres not going to be any use for the non full sized images. can we replace the largeimageUrl that comes with the image to be a fullsized jpg? or is it better to save the space, in which case I'll go with requesting a full size in a new feature

src/components/Project/ImageAnalysis/ImageDownloadMenu.vue
LTDakin: you made me have to do some freaky things to get this working but I learned something new

src/components/Project/ImageAnalysis/ImageDownloadMenu.vue
jnation3406: I would think we always want to have "large" but not full size images for the frontend use, since tablets don't have that high resolution and you don't need to see full rez on screen anyway... But if we did full rez everywhere it would be using a lot more bandwidth. You can run it past everyone else, but I think we should just have downloading the image in jpeg/tiff happen in the backend with a full size version, but keep the UI using the large not full size versions.

jnation3406 (approved):

PR #100, opened 2024-08-30
Shows the source catalog when loaded, Source catalog markers can't be edited

LTDakin:
Noticed that when people analyzed the image they tended to always turn on the source markers to get accurate lines between stars, to save them a click and improve the experience now when source catalog is loaded it is shown by default and can still be checked off.

Also fixed a bug where we allowed the editing of source catalog markers, now the layer can no longer be edited.

Screenshot 2024-08-30 at 3 06 47 PM


capetillo (approved):
nice work!

PR #101, opened 2024-09-03
rename session_id to camel case, loading bar button, immediate polling

LTDakin:
Improved the progress bar to be part of the button instead of disconnected underneath it
also renamed sessionId to be camel case to fix a vue linting error
we also now poll the button's operation status immediately and then set an interval so we can instantly update the state without having to wait 

old loading bar

https://github.com/user-attachments/assets/ece53212-553d-47f6-8fc1-8a43bfcf85dc

example with the button selected to show that it works with the selected color

https://github.com/user-attachments/assets/93dc2adf-d155-428b-bc15-a894f8210eb1

example with no selections

https://github.com/user-attachments/assets/fca03733-e8ce-4767-b2c3-5ee728f7cce4

### DIFF
--- a/src/components/Global/LoadBarButton.vue
+++ b/src/components/Global/LoadBarButton.vue
@@ -1,0 +1,44 @@
+<script setup>
+defineProps({
+  progress: {
+    type: Number,
+    required: true,
+  }
+})
+
+</script>
+
+<template>
+  <v-btn class="loadBarButton">
+    <slot />
+    <div
+      class="progress-bar"
+      :style="{ width: progress + '%' }"
+    />
+  </v-btn>
+</template>
+<style scoped>
+.loadBarButton{
+  position: relative;
+  overflow: hidden;
+  background-color: var(--dark-blue);
+}
+:slotted(p) {
+  z-index: 2;
+}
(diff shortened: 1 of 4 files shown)


src/components/Global/LoadBarButton.vue
capetillo, line 1: Will this component be used elsewhere? If not, then I think it makes sense to have this live in `OperationPipeline`. If you think it can be reused elsewhere then this work : )

capetillo (approved):
Looks awesome! One small comment. And if you think that component will be reused, then LGTM

src/components/Global/LoadBarButton.vue
capetillo: I meant it being reused as a component at all. It seems like the logic could be just defined in `OperationPipeline`. Sorry for the misunderstanding!

LTDakin: I like to treat components as classes, so they might be reused or just used once, but separating them has the advantage of keeping the parent class simple. So if we ever need to change the button in the future you can work with a tiny file with clear props, vs scrolling a large file that has multiple things in it and trying to figure out which variables are for what

PR #105, opened 2024-09-16
removed test images to reduce project size

LTDakin:
these were used during the skeleton phase of datalab when we didn't have archive data. cleaning them up


mgdaily (commented):
Nice. These will still be present in the git history, however, so we should probably remove them all from the git history as well. Otherwise they will still be present (though, hidden) every time someone clones the repository.

GitHub has a guide about this:

https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/removing-sensitive-data-from-a-repository

jnation3406 (approved):

LTDakin:
Putting this on pause until all other PRs are merged (required by BFG repo cleaner to properly clean the files)

PR #108, opened 2024-10-03
use attribute operation_progress due to name change in backend hackathon feedback PR

LTDakin:
Paired PR to the hackathon backend PR https://github.com/LCOGT/ptr_datalab/pull/35

### DIFF
--- a/src/components/DataSession/OperationPipeline.vue
+++ b/src/components/DataSession/OperationPipeline.vue
@@ -45,14 +45,11 @@ async function pollOperationCompletion(operationID) {
   // Success Callback for checking operation status
   const updateOperationStatus = (response) => {
     if(response){
-      const operationStatus = response.status
-      const percentCompletion = response.percent_completion
-
-      switch(operationStatus){
+      switch(response.status){
       case 'PENDING':
         break
       case 'IN_PROGRESS':
-        operationPercentages.value[operationID] = percentCompletion * DEC_TO_PERCENT
+        operationPercentages.value[operationID] = response.operation_progress * DEC_TO_PERCENT
         break
       case 'COMPLETED':
         operationPercentages.value[operationID] = COMPLETE_PERCENT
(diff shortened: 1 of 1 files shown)


capetillo (approved):
I think this looks good! However, and I may be wrong, but the `alertTimeStamp` seems a bit hacky. Is the `alertTimeStamp` supposed to function as some sort of unique id? And if so, is it possible to get two of those alerts at the exact same time? But again, I may be completely off and these questions stem from curiosity!

src/stores/alerts.js
```diff
@@ -8,16 +8,16 @@ import { defineStore } from 'pinia'
 export const useAlertsStore = defineStore('alerts', {
   state() {
     return{
-      // The type of alert to display ('success', 'error', 'warning', 'info')
-      alertType: 'warning',
-      // The text to display in the alert
-      alertText: 'default alert text',
+      alertType: 'default', // alert type to display ('success', 'error', 'warning', 'info', 'loading')
+      alertText: 'default alert text', // text displayed in alert
+      alertTimeStamp: '0' // timestamp of the last alert, used for watching in AlertToast.vue
```
jnation3406, line 13: Is there one global alert vs having the alert tied to the page it is alerting on because we want alerts to persist between different views? Otherwise I'm not sure why its in the global state/store instead of just being local to where it is shown...

jnation3406 (approved):
Just one question about the alerts store in general

src/stores/alerts.js
LTDakin: This is my argument for having it in a store but I could still be incorrect.
Persisting between views is one benefit, the other reason I had it in the store is because of its global nature. It's created once at the app level and done. Imported the alertToast into every single component that wants to show alerts is drawing multiple overlapping alertToast components. For example the leaflet map is in the analysis view which is in the project view. All of those components have the need to alert the user about status or errors. It feels nice just having two places to look when changing alerts, the store, the component, and you don't have to worry about anywhere else. Having it's state managed globally also allows us to easily create error logs if we want.

PR #109, opened 2024-10-11
Analysis data improvements

LTDakin:
Requested by Micheal to improve some aspects of the analysis view

- Site codes are now replaced with their full site name
- There is now an info button to view the Image Data (FITS headers)
- Other UI improvements

Screenshot 2024-10-11 at 1 20 41 PM

Screenshot 2024-11-01 at 10 41 27 AM

### DIFF
--- a/src/components/Project/ImageAnalysis/ImageAnalyzer.vue
+++ b/src/components/Project/ImageAnalysis/ImageAnalyzer.vue
@@ -6,6 +6,7 @@ import ImageViewer from './ImageViewer.vue'
 import LinePlot from './LinePlot.vue'
 import FilterBadge from '@/components/Global/FilterBadge.vue'
 import ImageDownloadMenu from '@/components/Project/ImageAnalysis/ImageDownloadMenu.vue'
+import { siteIDToName } from '@/utils/common'
 
 const props = defineProps({
   modelValue: {
@@ -27,6 +28,7 @@ const startCoords = ref()
 const endCoords = ref()
 const catalog = ref([])
 const positionAngle = ref()
+const showFitsHeaderDialog = ref(false)
 
 function closeDialog() {
   lineProfile.value = []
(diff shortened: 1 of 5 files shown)


src/components/Project/ImageAnalysis/ImageAnalyzer.vue
```diff
@@ -121,11 +128,49 @@ function handleAnalysisOutput(response, action, action_callback){
       </div>
     </v-sheet>
   </v-dialog>
+  <v-dialog
```
mgdaily, line 131: Unfortunately this isn't the FITS header, this is just the archive image record.

The FITS header data would need to either come from the backend via the FITS file, or the archive API. I think combining the headers from all HDUs in a FITS file and sending it alongside the large image URL would work.

You can also query it from the archive directly, e.g. `GET /frames/<id>/headers` but that wouldn't work if we decide to support images from other sources, like other archives.

src/utils/common.js
```diff
@@ -3,4 +3,19 @@ const calculateColumnSpan = (imageCount, imagesPerRow) => {
   return totalColumns
 }
 
-export { calculateColumnSpan }
+function siteIDToName(siteID) {
+  const siteIDMap = {
```
mgdaily, line 7: We may want to get this mapping from the backend and set it in the pinia store in the frontend. The backend could query ConfigDB and keep a basic mapping of sites to locations in the redis cache. That also makes it extensible in the future in case we want to add any additional information.

mgdaily (changes requested):
Mostly looks good! But there's a couple of things I think we should do differently.

src/components/Project/ImageAnalysis/ImageAnalyzer.vue
LTDakin: Oops, it had some similar data so I thought it was the header but I can see this is archive specific now.

src/utils/common.js
LTDakin: can't we have the Frontend query ConfigDB directly and store it in Pinia? That seems more direct and more easily extensible

src/components/Project/ImageAnalysis/ImageAnalyzer.vue
LTDakin: This is getting into multiple archive support, I think there should be some planning around this before I just come up with a temporary solution that may not work when we actually implement it

mgdaily: Right - agreed that we will need to plan for multiple archive support, but returning the header data from the backend will work regardless of what the data source is - we just need the FITS file and we can serve it.

src/utils/common.js
mgdaily: Unfortunately ConfigDB is hosted within the LCO internal VPN, so when a client outside of the VPN or LCO building wants to query configDB, it needs to happen via a service that runs inside the VPN.

src/components/Project/ImageAnalysis/ImageAnalyzer.vue
LTDakin: I think in that case then I need to make a new header endpoint because we arn't getting the image URLs from the backend *I think* The way Jon wrote the image caching is that it goes to the archive directly?

mgdaily (approved):

PR #111, opened 2024-10-30
Feature/rgb image scaling

jnation3406:
Added an extra page to get scaling parameters for input images in the OperationWizard. Also made adding the operation contingent on setting the required number of parameters in the wizard. I also attempted to fix some of the random errors we would get when caching jpegs that couldn't be found.

### DIFF
--- a/public/drawImageWorker.js
+++ b/public/drawImageWorker.js
@@ -1,0 +1,46 @@
+
+var context, width, height, imageData, imageDataArray, sharedArrayBuffer, sharedArray, gammaTable
+ 
+// Web workers function a little differently in that this method is always called
+// from the main thread, but it can persist values in the worker, and what it does
+// is based on what arguments are passed in from the main thread.
+// This lets us seed the web worker with some initial data that it uses
+// to recompute the pixel values with each slider value change
+onmessage = function(event) {
+  if (event.data.canvas) {
+    // This is the starter message with the canvas, width, height, imageData, and shared data array
+    context = event.data.canvas.getContext('2d')
+    width = event.data.width
+    height = event.data.height
+    sharedArrayBuffer = event.data.sharedArrayBuffer
+    sharedArray = new Uint8ClampedArray(sharedArrayBuffer)
+    imageDataObject = context.createImageData(width, height)
+    let size = 256
+    let gamma = 2.5
+    gammaTable = []
+    for (let i = 0; i < size; i++) {
+      gammaTable.push(parseInt(size * Math.pow(i / size, 1.0 / gamma)))
+    }
+  }
+  if (event.data.imageData) {
+    imageData = event.data.imageData
+  }
+  if (event.data.scalePoints) {
+    // If scalePoints is sent, then we can re-compute the image and redraw it to the canvas
+    const low16Bit = parseInt(event.data.scalePoints[0])
+    const high16Bit = parseInt(event.data.scalePoints[1])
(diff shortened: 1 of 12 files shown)


public/drawImageWorker.js
```diff
@@ -0,0 +1,46 @@
+
```
jnation3406, line 1: Web workers need to be somewhere accessible for the frontend main thread to access them during exection, so this public dir works.

src/utils/common.js
```diff
@@ -3,4 +3,28 @@ const calculateColumnSpan = (imageCount, imagesPerRow) => {
   return totalColumns
 }
 
-export { calculateColumnSpan }
+const convertFilter = (filter) => {
+  if (filter === 'V' || filter === 'gp') {
```
jnation3406, line 7: This function maps the potential filters into 'r', 'g', 'b'. We can expand this as we accept more filter names to mean those colors.

vue.config.js
```diff
@@ -6,5 +6,11 @@ module.exports = defineConfig({
 		vuetify: {
 			// https://github.com/vuetifyjs/vuetify-loader/tree/next/packages/vuetify-loader
 		}
+	},
+	devServer: {
```
jnation3406, line 10: This is necessary for your locally running devServer to be able to use web workers with shared data arrays. It requires https and same origin normally, so I think it should work in production but we may need to set some cors thing for that as well.

src/components/Global/ImageScaler.vue
```diff
@@ -0,0 +1,150 @@
...
+      rawData.value = response
+      zScaleValues = [response.zmin, response.zmax]
+      zScaleImage()
+      isLoading.value = false
+    },
+    failCallback: (error) => {
+      isLoading.value = false
+      errorReason.value = error
+      console.error('API call failed with error:', error)
+    }
```
LTDakin, line 78: this is a very readable way to use lamdas for the fetchapi method, I'll be stealing this

LTDakin (commented):
I gave it a look through and it looks really good, I think @mgdaily should give the final approval since he initially prototyped this

src/components/Global/ImageScaler.vue
```diff
@@ -0,0 +1,150 @@
...
+  }
+})
+
+const emit = defineEmits(['updateScaling'])
+const store = useConfigurationStore()
+const dataSessionsUrl = store.datalabApiBaseUrl
+const isLoading = ref(true)
+const errorReason = ref('')
+const rawData = ref({})
+const sliderRange = ref([0, 65535])
```
mgdaily, line 35: Maybe we add a TODO to make this work with an arbitrary bit depth? The site "simple-stacked" images will be 32 bit.

src/components/Global/ImageScaler.vue
```diff
@@ -0,0 +1,150 @@
...
+        :image-name="props.imageName"
+        :composite-name="props.compositeName"
+      >
+      </raw-scaled-image>
+    </v-row>
+    <v-row>
+      <v-range-slider
+        v-model="sliderRange"
+        step="500"
+        max="65535"
```
mgdaily, line 102: Again, may want to make this computed based on the bit depth of the image.

mgdaily (approved):
Mostly looks good! Just some questions about bit depth.

src/components/Global/ImageScaler.vue
jnation3406: hooked in the bitdepth

PR #114, opened 2024-11-05
moved click event to the title instead of panel

LTDakin:
Problem: Clicking on the background of the grid or empty space that was part of the expansion panel the images would perform a reload when they didn't need to. 
Solution: This is now fixed by moving the click event to the title bar so only that will reload the page.
Cause: Not sure, as this code wasn't changed recently and it worked before. Maybe Vuetify changed something about the bounds of expansion panels

Before Fix:

https://github.com/user-attachments/assets/743aceff-a306-4391-b5f0-c980a0253972

After Fix:

https://github.com/user-attachments/assets/00dfdfb5-6234-47d5-ae6f-524062157d34

### DIFF
--- a/src/views/ProjectView.vue
+++ b/src/views/ProjectView.vue
@@ -224,9 +224,8 @@ onMounted(() => {
       <v-expansion-panel
         v-for="proposal in userDataStore.proposals"
         :key="proposal.id"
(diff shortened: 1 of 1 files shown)


mgdaily (approved):

capetillo (approved):
nice!

PR #119, opened 2024-11-08
fix hydration error for vue config

LTDakin:
Fixes a hydration warning that wouldn't break the website, but it does tell Vue which packages to include at runtime. Without it we would be including some dev tools packages by default so this helps reduce the build size and overhead.

Screenshot 2024-11-08 at 11 21 41 AM

In the above image I ran the old code which shows the warning, then I stopped the server, ran it with the options set, and the warning is gone.


mgdaily (approved):

jnation3406 (approved):
Looks fine

PR #123, opened 2024-11-12
Fix/data session view behavior fixes

LTDakin:
A list of Bug fixes and behavior fixes on the data sessions view

- fixed `missing data sessions` error message by adding it into a check if the data sessions list is 0
- v-content deprecated printing a warning to console, replaced with v-main
- linting
- removed class from image grid since it wasn't applying to anything, image grid has multiple roots
- fixed behavior where you would delete operation 1 and it would delete but then select the new operation 1. Now the selected operation index is reset when an operation is deleted.
- New `deleteOperation` api util function to reduce repeated code, used multiple places
- Removed an old error snackbar for delete operation failures, any api errors are handled by the global error alert store

### DIFF
--- a/src/components/DataSession/OperationPipeline.vue
+++ b/src/components/DataSession/OperationPipeline.vue
@@ -90,6 +90,11 @@ function openDeleteOperationDialog(operation) {
   showDeleteDialog.value = true
 }
 
+function itemDeleted(){
+  selectedOperation.value = -1
+  emit('operationWasDeleted')
+}
+
 watch(() => props.operations, () => {
   if (props.active) {
     props.operations.forEach(operation => {
@@ -167,7 +172,7 @@ onBeforeUnmount(() => {
     v-model="showDeleteDialog"
     :session-id="sessionId"
     :operation-id="deleteOperationId"
-    @item-was-deleted="emit('operationWasDeleted', $event)"
+    @item-was-deleted="itemDeleted()"
   />
 </template>
 
@@ -185,9 +190,8 @@ onBeforeUnmount(() => {
 .operation_button {
   width: 12rem;
   height: 3rem;
-  font-size: 1.2rem;
+  font-size: 1rem;
   font-weight: 600;
-  border-style: none;
   color: var(--metal);
 }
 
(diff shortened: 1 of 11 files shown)


src/utils/api.js
```diff
@@ -46,5 +47,11 @@ const handleError = (error) => {
   console.error('API call failed with error:', error)
 }
 
+function deleteOperation(sessionId, operationId, successCallback = null, failCallback = handleError) {
```
capetillo, line 50: Is `successCallback = null` needed?

src/views/DataSessionsView.vue
```diff
@@ -11,25 +11,21 @@ const router = useRouter()
...
 const dataSessionsUrl = configurationStore.datalabApiBaseUrl + 'datasessions/'
 
 onMounted(async () => {
   await loadSessions()
-  if (dataSessions.value.length > 0 && !userDataStore.mostRecentSessionId) {
-    const firstSessionId = dataSessions.value[0].id
-    tab.value = firstSessionId
-    userDataStore.mostRecentSessionId = firstSessionId
-  } else {
+  if (dataSessions.value.length > 0 && !userDataStore.activeSessionId) {
```
capetillo, line 20: It would be useful to add a comment here explaining what this is doing : )

src/views/DataSessionsView.vue
```diff
@@ -58,7 +54,7 @@ function tabColor(index) {
 }
 
 function tabActive(index) {
-  if (dataSessions.value[index] && tab.value === dataSessions.value[index].id) {
+  if (dataSessions.value[index] && userDataStore.activeSessionId === dataSessions.value[index].id) {
```
capetillo, line 57: also here though I suspect it has to do with which tab is active but still not sure what it's doing in the UI

capetillo (commented):
Left a comment and a couple of questions

src/components/DataSession/OperationPipeline.vue
```diff
@@ -90,6 +90,11 @@ function openDeleteOperationDialog(operation) {
   showDeleteDialog.value = true
 }
 
+function itemDeleted(){
+  selectedOperation.value = -1
```
mgdaily, line 94: Perhaps leave a comment explaining why we're setting this to -1?

mgdaily (approved):
Looks good, I agree with Carolina's comments!

PR #126, opened 2024-11-14
Broken slider for v-tabs and routerlink

LTDakin:
slider for vuetify was not properly selecting the active tab, changed it to use the routerlink class selector with a matching background color to content

Post Fix

https://github.com/user-attachments/assets/39fceec4-7482-42ec-8e93-fe2f41e40edc

Pre Fix

https://github.com/user-attachments/assets/cf75aa16-8256-4ba6-84ad-3b2035199617


jnation3406 (approved):
Nice

PR #127, opened 2024-11-16
Feature/histogram slider

jnation3406:
This adds the HistogramSlider component and uses that for the image scaling controls. It depends on [this backend PR](https://github.com/LCOGT/ptr_datalab/pull/41).

[histogram_scaling.webm](https://github.com/user-attachments/assets/d2c08027-d4cd-4fae-b487-3c1f8ed2cc5b)


src/components/Global/HistogramSlider.vue
```diff
@@ -0,0 +1,229 @@
...
+        track-size="0"
+        :track-color="backgroundColor"
+        :track-fill-color="selectedColor"
+        :thumb-color="selectedColor"
+        thumb-size="16"
+        :max="props.bins.length-1"
+        thumb-label="always"
+        strict
+        hide-details
+        @update:modelValue="updateScaleRange"
```
LTDakin, line 205: Don't need to change it, but wanted to check this to help me learn more about vue. A watch function on slider Range would be equivalent to listening for an update event of the components model Value. Only benefit is someone reading it that doesn't know vue as well may be able to understand watchSliderRange vs making the connection that the variable in v-model is what is triggering the @ update:modelValue event.

LTDakin (approved):
Looks good, just some file organization

src/components/Global/HistogramSlider.vue
jnation3406: So there can actually be a difference in behavior between those two things in practice. Setting an on update handler in the component will only get triggered when the value changes due to the component, i.e. by moving the slider. This is different from a watcher, that will get triggered any time the value changes by any means, including just setting it elsewhere in the code. So you'll see I do slightly different logic based on if the value is changed from the range slider vs from the number inputs, even though both of those things will ultimately change the value in the variable.

mgdaily (approved):

PR #130, opened 2024-12-02
Fixes for datalab output analysis

LTDakin:
Problem: when trying to do a line profile, fetch source catalog, or download fits files for datalab outputs you would get errors or features would be missing.

Packages: 
- removed the `vue3-carousel` package since its no longer used
- replaced `d3` with `chart js`, reasoning for this is d3 is a very custom graphics package that trades more versatility for harder maintenance. Replaced with the charting library chart js as it offers nice features out of the box and is more suited for our simple use case.

CreateSessionDialog.vue
- `id` passed so calls to the archive for fits headers works in data session view
- `url` passed so download of fits files works in data session view

ImageAnalyzer.vue
- now correctly clears `position angle`
- `source` is passed to back-end so that analysis actions can work for Datalab outputs

ImageDownloadMenu.vue
- instead of passing the whole image object, just passes the attributes needed

ImageViewer.vue
- move creation of layer control to the success callback of source catalog so that an empty menu is not drawn when there is no source catalog

LinePlot.vue
- replaced d3 line chart with, hopefully, much easier to read/edit chart js

Api.js
- api fail callback now checks response for an error message from the backend, if it has one it creates an alert, otherwise it logs the error to the console.

https://github.com/user-attachments/assets/903f7a9c-1e8e-4e48-9f01-4f2a646b589f

Screenshot 2024-12-02 at 11 48 50 AM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,8 +10,8 @@
       "dependencies": {
         "@geoman-io/leaflet-geoman-free": "^2.17.0",
         "@mdi/font": "5.9.55",
+        "chart.js": "^4.4.6",
         "core-js": "^3.8.3",
-        "d3": "^7.9.0",
         "leaflet": "^1.9.4",
         "lodash": "^4.17.21",
         "lru-cache": "^10.2.0",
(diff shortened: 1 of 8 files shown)


jnation3406 (approved):
Looks good to me. Overall I think its good to not use d3 when a higher level library can do what you need, but that being said I might have to re-add d3 if we go to doing the graph representation of operations.

PR #138, opened 2025-01-10
UI refresh of the RGB stack component

LTDakin:
RGB stacking UI refresh to fit the controls on one screen. Doesn't change any of the functionality of how RGB stacks work.

Updated UI
Screenshot 2025-01-09 at 3 53 02 PM

Previous UI

image

### DIFF
--- a/src/components/Global/Scaling/CompositeImage.vue
+++ b/src/components/Global/Scaling/CompositeImage.vue
@@ -1,6 +1,7 @@
 <script setup>
 import { ref, defineProps, computed, watch, onMounted } from 'vue'
 import { useScalingStore } from '@/stores/scaling'
+import { scaleImageData } from '@/utils/common'
 
 // This component draws a composite RGB image from the scaling store
 
@@ -32,7 +33,9 @@ const ableToDraw = computed(() => {
 
 function redrawImage() {
   if (ableToDraw.value) {
-    context.putImageData(store.scaledImageArrays[props.imageName].combined, 0, 0)
+    const compositeImage = store.scaledImageArrays[props.imageName].combined
+    const scaledCompositeImage = scaleImageData(compositeImage, props.width, props.height)
+    context.putImageData(scaledCompositeImage, 0, 0)
   }
 }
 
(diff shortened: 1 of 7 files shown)


src/components/Global/Scaling/CompositeImage.vue
```diff
@@ -32,7 +33,9 @@ const ableToDraw = computed(() => {
 
 function redrawImage() {
   if (ableToDraw.value) {
-    context.putImageData(store.scaledImageArrays[props.imageName].combined, 0, 0)
+    const compositeImage = store.scaledImageArrays[props.imageName].combined
+    const scaledCompositeImage = scaleImageData(compositeImage, props.width, props.height)
```
jnation3406, line 37: I'm guessing this isn't super performant since it makes another context and draws another image, but its probably okay since you shrunk the 3 raw images so overall went from 4 big images to 2 (for scaling this)...

src/components/Global/Scaling/RawScaledImage.vue
```diff
@@ -78,9 +78,18 @@ watch(
...
+    <canvas
+      ref="imageCanvas"
+      class="raw-scaled-canvas"
+      :width="props.maxSize"
+      :height="props.maxSize"
+    />
   </div>
 </template>
 <style scoped>
+.raw-scaled-canvas {
```
jnation3406, line 91: is this just going to override the width and height setting of the canvas from props.maxSize to 200px?

jnation3406 (approved):
I'm a little worried about the performance hit from the scaling, since that is basically "drawing" 2 extra images, but its probably masked from the performance gains of having the 3 raw-images shrunk down... As long as its still pretty much real-time responsive i guess its fine.

src/components/Global/Scaling/RawScaledImage.vue
LTDakin: thats a weird canvas quirk. So they're drawn at 500px quality wise (which impacts the quality of the composite image) and the css shrinks them down when displaying them. I realized this may be a solution for the large image as well to reduce the performance cost of re-sizing it.

src/components/Global/Scaling/CompositeImage.vue
LTDakin: oops actually I think this might just worsen performance. I'm only scaling the composite image and not the 3 smaller ones. I was originally using this code when I shrank the raw images to 200 px and had to increase the size of the final composite image. But now they're still 500px so I may be able to get rid of this and just use css to size the canvas

LTDakin: after some digging in the MDN web docs I found that `ctx.drawImage` has additional parameters for scaling but requires the data to be in `imageBitMap` form. So I used the function `createImageBitmap` to convert `imageData` to `ImageBitMap`. Removed this extra canvas drawing function

PR #139, opened 2025-01-16
Added a graph view of the operations using vueflow

jnation3406:
This adds a graph view of the operations into the DataSession UI. I takes up the full datasession viewport since it can get big and messy if you have many operations, and you can toggle to it with a button next to the Operations, and then go back by closing it. I experimented with having images in the graph with the ImagesNode I committed but ultimately left it out because I struggled to get the images sized and positioned such that it looked good.

The selection of the operation is linked between the two operation views. In the future we may want to add the delete button to the graph view as well and unify the delete dialog - Also deleting should delete all operations that depend on that operation in a chain, which should be made visible to the user in the delete dialog before we delete as well. Currently it just deletes the one operation and could break future operations that depended on it I think . We could also unify the polling of operations between the graph and linear view, and then the graph view could show something happening based on the loading status.

Some known issues:
- The container the graph is drawn into is hardcoded in css to have 800px height. Leaving out the hardcoded prevents the graph from being able to layout/draw because it thinks there is no room. Don't what combination of CSS could fix this, or maybe we can just set it to the screen window size instead of 800px or something
- I'm currently using the LoadBarButtons as the nodes, and there's some weird behavior where they get darker as you click them and then lighten as you hover them. Some interaction of the button with VueFlow nodes I guess. Since I'm not even using the loading bar part and just want the styling, we could change this to be a normal button styled the same.

 
[operation_graph.webm](https://github.com/user-attachments/assets/51fa690d-a53f-4139-a66d-f02d35923a4a)


LTDakin (approved):
Looks really nice even with the limitations you mentioned! Thanks for spending a lot of time exploring this, it will make it much easier to approach the idea again in the future I'm sure. I'm going to hold off on merging this just so that when I have some time I can run it and make a few css changes like the black border around the data sessions view.

LTDakin:
- Removed extra gray padding that was around the Data Session
- Changed the color of the back button and icon to be seen easier 
- added a background to the graph

Screenshot 2025-01-21 at 1 09 57 PM

PR #140, opened 2025-01-23
Toggle Image Sources with Icon instead of leaflet menu

LTDakin:
Small improvement to consolidate controlling the source catalog with the icon next to the range slider. This will be more intuitive to the user as ot follows better UX principles by co-locating related functions. 

Also moved the image info back to the top

Old position of source catalog control
Screenshot 2025-01-22 at 4 04 17 PM

New co-located control

https://github.com/user-attachments/assets/6b77b30e-c016-494d-b4d9-f135f63f366c

### DIFF
--- a/src/components/Project/ImageAnalysis/ImageAnalyzer.vue
+++ b/src/components/Project/ImageAnalysis/ImageAnalyzer.vue
@@ -29,6 +29,7 @@ const lineProfile = ref([])
 const lineProfileLength = ref()
 const startCoords = ref()
 const endCoords = ref()
+const catalogToggle = ref(true)
 const catalog = ref([])
 const positionAngle = ref()
 const headerDialog = ref(false)
@@ -36,7 +37,12 @@ const headerData = ref({})
 const fluxSliderRange = ref([0, 10000])
 
 const filteredCatalog = computed(() => {
-  return catalog.value.filter((source) => source.flux >= fluxSliderRange.value[0] && source.flux <= fluxSliderRange.value[1])
+  if(catalogToggle.value){
+    return catalog.value.filter((source) => source.flux >= fluxSliderRange.value[0] && source.flux <= fluxSliderRange.value[1])
+  }
+  else{
+    return []
+  }
 })
 
 function closeDialog() {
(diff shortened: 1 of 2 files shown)


src/components/Project/ImageAnalysis/ImageAnalyzer.vue
```diff
@@ -142,21 +148,46 @@ function showHeaderDialog() {
...
           >
             <b>{{ filteredCatalog.length }} Sources with Flux between {{ fluxSliderRange[0] }} and {{ fluxSliderRange[1] }}</b>
-            <non-linear-slider
-              v-model="fluxSliderRange"
-              prepend-icon="mdi-flare"
-              :max="Math.max(...catalog.map((source) => source.flux))"
-              :min="Math.min(...catalog.map((source) => source.flux))"
-            />
+            <div class="d-flex justify-end">
+              <v-icon
```
jnation3406, line 176: Two things I think would improve this a bit: But the icon inside a button so its more clear when its turned on/off, and also adding a tooltip to the icon/button that says "Toggle Sourcemap" or something like that.

jnation3406 (approved):
Looks better, just some suggestions to make it even more clear.

src/components/Project/ImageAnalysis/ImageAnalyzer.vue
LTDakin: Screenshot 2025-01-23 at 10 15 14 AM

capetillo (commented):
I'm not sure if in the screen recording you clicked and then the `fluxSliderRange` appeared. If so, I don't think that's intuitive design. Otherwise ignore this comment. Also, can you add comments to your code so that when another engineer (e.g. myself) works on this, they know what they're working on without having to read a lot and decipher : )

src/components/Project/ImageAnalysis/ImageAnalyzer.vue
```diff
@@ -29,14 +29,20 @@ const lineProfile = ref([])
...
+const catalogToggle = ref(true)
 const catalog = ref([])
 const positionAngle = ref()
 const headerDialog = ref(false)
 const headerData = ref({})
 const fluxSliderRange = ref([0, 10000])
 
 const filteredCatalog = computed(() => {
-  return catalog.value.filter((source) => source.flux >= fluxSliderRange.value[0] && source.flux <= fluxSliderRange.value[1])
+  if(catalogToggle.value){
```
capetillo, line 40: doesn't `catalogToggle` always have a value, even if that value is `false`? It's initialized as true

LTDakin: To answer your above, that pop in was the source catalog loading, maybe in the future I could animate it sliding down out of the image info so its smoother

As for this comment I'm a little confused, `catalogToggle.value` will always have a value either `true` or `false` for the if condition

PR #141, opened 2025-01-24
fixes one way difference bug, now checks symmetrically

LTDakin:
Bug: In the list view checking an image would select it to be added to a proposal but unchecking it wouldn't

Cause: I wasn't checking the [symmetrical difference](https://stackoverflow.com/questions/1187518/how-to-get-the-difference-between-two-arrays-in-javascript) between the sets and so it would only find the difference between the table and proposal but not vice versa

Fix: concat the includes check both ways to find the total difference


jnation3406 (approved):
I don't really understand why we have this - isn't this getting the images not selected in either mode? why would we want that? But I trust it does whatever you wanted it to do...

PR #153, opened 2025-02-06
base scaling off input filter, remove check for correct filter to allow any

LTDakin:
This change allows for any filter to be added as input for red green and blue in the RGB stack operation.
To do this a filtering step was removed as well as passing down the input type to determine the filter instead of the image's filter.

Also changed the switch case to use a map instead and removed observatories from the text conversion as it was too long inside the table view, but that change is unrelated and can be reverted if need be.

[Backend change to allow non aligned images to be stacked ](https://github.com/LCOGT/ptr_datalab/pull/55) is needed to make this work as well

https://github.com/user-attachments/assets/22f0588d-2af0-4373-bbb7-d68b5b1315c0

### DIFF
--- a/src/components/DataSession/OperationWizard.vue
+++ b/src/components/DataSession/OperationWizard.vue
@@ -140,7 +140,7 @@ function goForward() {
     // if there are no images for a filter required by the operation, do not proceed
     for (const inputKey in selectedOperationInputs.value) {
       const inputField = selectedOperationInputs.value[inputKey]
-      if (inputField.type == 'file' && imagesWithFilter(inputField.filter).length == 0){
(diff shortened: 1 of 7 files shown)


jnation3406 (commented):
Looks fine, but it would be nice to sort the input images by the preferred colors first. So for example, all images are available to select for R, G, B channels, but the order they are presented to you is different for each channel, with the r, R, rp filter images shown first in the R channel and so on. This won't matter much when you have like 5 images in your session, but when you start having 10+ it'll help users see and pick the most common filters for those channels first.

jnation3406 (approved):
Looks good, and good idea referencing by basename. I'm switching the operations code to reference by operation id instead of index as well for similar reasons.

PR #154, opened 2025-02-12
Add in bulk deletion of operations, so any dependent operations are d…

jnation3406:
- Add bulk deletion of operations to delete any chained dependent operations with it
- Refactored some of the code so `index` and `dependencies` are calculated and stored in the operations so they can be reused within components. 
- Changed a bunch of internal code for selection and stuff to use operation id instead of operation index - index should just be shown on the UI to the user but not used internally anywhere else
- Changed the way the node graph was deciding on dependencies so it doesn't use the output of the operation, which didn't work until the operation was complete. Now it uses the operations input, which contains the operation id associated with any input files that come from datalab.

[bulk_delete.webm](https://github.com/user-attachments/assets/85796f92-5628-4881-9092-5e986d8a8c34)


src/components/DataSession/DataSession.vue
```diff
@@ -117,19 +115,34 @@ function reconcileOperations() {
...
-    selectedOperation = -1
+  if (!(selectedOperation.value in operationMap)) {
+    selectedOperation.value = -1
   }
   reconcileOperationImages()
 }
 
 watch(
   () => props.data.operations, () => {
+    // Look through the input_data for file arrays and set dependency set on each operation
```
LTDakin, line 128: Can this be seperated into a method called addOpDependencies or whatever name you'd like. I'm trying to keep any Lifecycle hooks (onMounted, watch, etc.) simple where they can be. Makes debugging orders of things happening on the browser easier

src/components/DataSession/OperationGraph/OperationPipelineFlow.vue
```diff
@@ -67,50 +66,26 @@ function updateNodesAndEdges() {
...
-                  type: 'smoothstep',
-                  ...getEdgeProps(outputImageToOperation[inputValue.basename].toString(), index.toString())
-                })
-                // A Set of edgeIds is used to make sure we don't double create edges
-                edgeIds.add(edgeId)
-              }
-            }
-          })
-        }
+    props.operations.forEach((operation) => {
```
LTDakin, line 79: wow really nice improvement with operation Ids instead of indexes.

src/utils/api.js
```diff
@@ -54,5 +54,12 @@ function deleteOperation(sessionId, operationId, successCallback, failCallback =
   fetchApiCall({ url: url, method: 'DELETE', successCallback: successCallback, failCallback: failCallback })
 }
 
+function deleteOperations(sessionId, operationIds, successCallback, failCallback = handleError) {
```
LTDakin, line 57: This is a good practice I have to remember to do more with the code. More specific fetch calls in the api util file that line up with the backend api. So when we hit one endpoint from many places in the frontend they're all using the same logic to reduce errors.

LTDakin (approved):
Just one change to refactor some code out of a lifecylce hook into its own function. Otherwise looks good

PR #155, opened 2025-02-12
Update/analysis file org

LTDakin:
Moving analysis to be a view since it's becoming a main part of the data lab ecosystem as well as helps to separate it from the components that make it up. Move those components to be grouped into their own analysis folder similar to Project and Data session.

Also updated the filter badge appearance and removed the two colors in preparation for a standard PTR styling


jnation3406 (approved):
Looks fine, did you just remove the colorblind mode though or is it moved?

PR #157, opened 2025-02-18
Move operation polling to the DataSession.vue and clean it up to work…

jnation3406:
… in graph dependency order. Also add some display things to the operation buttons to see them when pending, in progress, failed, and completed.

I changed the text color of pending operations to be white since its on a black background, so you can see what they are. I also changed the text color of in progress operations to be a moving gradient - this is to show it is in progress (the movement), and also so the text shows up on both the black and white backgrounds of the progress bar (the color). I've also removed the auto-deletion of failed operations, since they would just disappear too quickly for the user to even process what the error was and learn from it. Now they show up with a red background, and must be manually deleted - the user can see the error message as a tooltip by hovering over them to try to figure out their (or our) mistake.

[retry_operations.webm](https://github.com/user-attachments/assets/0a44e93a-9110-4815-b32b-38b801225dfe)


src/components/DataSession/DataSession.vue
```diff
@@ -142,16 +109,101 @@ function processOperations() {
   })
 }
 
-watch(
-  () => props.data.operations, () => {
+function clearPolling(operationID) {
```
LTDakin, line 112: to differentiate this from stopOperationPolling can we rename this one to stopPollingId(operationId) or something like that to indicate its for just one operation.

LTDakin (approved):
Really nice pulling the operation management outside of the pipeline, makes a lot more sense since DataSession.vue was already managing the state. Just some comments on naming/small changes. Up to you if you want to implement them.

PR #158, opened 2025-02-24
Unifying styles for all PTR sites

zemogle:
I've made some edits for the styles. I've not changed the layout with the exception of the Navbar. Vuetify doesn't really have a navbar type, just a toolbar which gets a bit tricky to use for navigation. I've added Bulma for handling this. It's classes a well namespaced so it won't clash with any Vuetify classes. Bulma is also light - 700Kb and adds both grid and column layouts if we want them.

The rest of this just changes colours. I've also changed the buttons on the operations screens to have coloured backgrounds (and look more like buttons). I found they weren't very visible when it was just using coloured text.

You'll notice that I've put the full PhotoRanch λ logo in the navbar. The title of the site, ie. "Datalab" is a menu to show which mini-site you're on, and switch between them.

Because there was more than just colours in the css file, I've renamed it (to `ptr-extra.css`) and updated the reference in `main.js`

I've done a basic check on a tablet too.


capetillo (approved):
Yay, nice work Edward!

LTDakin (approved):
Thanks for the styling it looks great. @jnation3406 it seems like he changed the operation button styling the same way you did to make it more readable. Will merge this in first and just be aware you might have to resolve conflicts

PR #159, opened 2025-02-25
Feature/analysis histogram

LTDakin:


src/views/AnalysisView.vue
```diff
@@ -46,16 +48,16 @@ const filteredCatalog = computed(() => {
...
-  emit('update:modelValue', false)
-}
+onMounted(() => {
+  analysisStore.image = props.image
+  analysisStore.imageUrl = props.image.largeCachedUrl
+  instantiateScalerWorker()
+})
+
+onUnmounted(() => {
+  imgWorker.terminate()
```
jnation3406, line 58: We never call this in the RawScaledImage component, should we add this there too?

src/views/AnalysisView.vue
```diff
@@ -110,66 +112,94 @@ function showHeaderDialog() {
...
+      await analysisStore.loadScaleData()
+    } 
+    catch (error) {
+      console.error('Failed to load scale data:', error)
+      return
+    }
+  }
+
+  // Create a new offscreen canvas and shared array buffer for the worker
+  const imgScalingCanvas = document.createElement('canvas')
```
jnation3406, line 129: If you are not actually showing the the canvas, then wouldn't it make sense to remove the offscreen canvas stuff from your web worker like we talked about before? I.e. use a different web worker from the RawScaledImage which does display the canvas image and cut out that part, and only do the part where it sets the shared memory in the array?

src/views/AnalysisView.vue
```diff
@@ -110,66 +112,94 @@ function showHeaderDialog() {
...
+  imgScalingCanvas.height = analysisStore.imageHeight
+  const offscreen = imgScalingCanvas.transferControlToOffscreen()
+
+  // Send the offscreen canvas and raw image data to the worker
+  imgWorker.postMessage({canvas: offscreen, width: analysisStore.imageWidth, height: analysisStore.imageHeight}, [offscreen])
+  imgWorker.postMessage({imageData: structuredClone(analysisStore.rawData)})
+
+  // Callback function for when worker completes, updates leaflet image
+  imgWorker.onmessage = () => {
+    analysisStore.imageUrl = imgScalingCanvas.toDataURL()
```
jnation3406, line 140: Oh I see you are pulling the image out of the offscreen canvas here... I don't actually know if that is faster or slower then just using the shared memory array to write the pixels into whatever your leaflet thing displays. But if this is the fastest way, then your drawImageWorker doesn't need to use the shared memory array at all, right?

src/utils/common.js
```diff
@@ -53,4 +53,14 @@ function initializeDate(dateString = 'none', defaultOffsetDays = 0) {
   return isNaN(date.getTime()) ? new Date(Date.now() + defaultOffsetDays * 24 * 3600 * 1000) : date
 }
 
-export { calculateColumnSpan, siteIDToName, initializeDate, filterToPixelIndex, filterToColor }
+// Utility function to load an image
+function loadImage(src) {
```
jnation3406, line 57: What is this for?

src/components/Analysis/ImageViewer.vue
```diff
@@ -21,69 +19,83 @@ const props = defineProps({
...
+    updateImageOverlay(newImageUrl)
+  }
+}
+)
 
-    // source catalog requires image dimensions to be fetched
-    fetchCatalog()
+// Loads image overlay and sets bounds
+async function loadImageOverlay(imgSrc) {
+  const img = await loadImage(imgSrc)
```
jnation3406, line 52: So your calling async functions but then just await-ing stuff in them, which is blocking right? Why not just let loading the image be asynchronous and then setting the width/height, and basename on load?

src/components/Global/Scaling/HistogramSlider.vue
```diff
@@ -67,15 +67,6 @@ function sliderToBinValue(sliderValue) {
   }
 }
 
-function labelSliderToScaleValue(sliderValue) {
```
jnation3406, line 70: Why did you get rid of this? This seems really important or else you aren't showing the correct values for the slider label - remember our histogram is using non-linear bins.

src/components/Global/Scaling/ImageScaler.vue
```diff
@@ -16,26 +17,30 @@ const props = defineProps({
   },
   maxSize: {
     type: Number,
-    default: 500
+    default: 700
```
jnation3406, line 20: This will affect the ImageScalingGroup used in the RGB operation selection

src/components/Global/Scaling/ImageScaler.vue
```diff
@@ -16,26 +17,30 @@ const props = defineProps({
...
+  },
+  showSample: {
+    type: Boolean,
+    default: true
   }
 })
 
 const emit = defineEmits(['updateScaling'])
 const store = useConfigurationStore()
+const analysisStore = useAnalysisStore()
```
jnation3406, line 38: Using the analysisStore doesn't make sense in this component right? How will this possibly work with the RGB operation selection since there are 3 of these ImageScaler instances shown at a time? It seems like you modified things to work for the AnalysisView but broke things for the RGB use case then...

src/stores/scaling.js
```diff
@@ -15,21 +15,29 @@ export const useScalingStore = defineStore('scaling', {
...
+        combined.data.fill(255)
+        scaledImages[combinedImageName]['combined'] = combined
       }
 
       const filterIndex = filterToPixelIndex(filter)
+      const combinedData = combined.data
 
       for (let i = filterIndex, j=0; j < imageDataArray.length; i += 4, j++) {
-        this.scaledImageArrays[combinedImageName]['combined'].data[i] = imageDataArray[j]
+        combinedData[i] = imageDataArray[j]
```
jnation3406, line 38: Ah so this is probably the indexing optimization you were talking about... makes sense.

jnation3406 (changes requested):
Most of it looks good but I think you may have broken the RGB image operation stuff in trying to get this to work with the same components...

src/components/Global/Scaling/ImageScaler.vue
LTDakin: yeah I upped the resolution for the rgb operation preview because of the optimizations so it can look nicer, I can revert it though

src/components/Analysis/ImageViewer.vue
LTDakin: The width and height properties of the Image() class are only available after its finished loading

src/utils/common.js
LTDakin: Loading images seems to be a pretty universal operation so this isn't really necessary but its just to save some code in the future and make it read nicer, so instead of 
```
const img = new Image(src)
img.onload( () => {
  // do stuff
}
```
you can write
```
const img = await loadImage(src)
```
Can remove it though and use the onload callback too

src/views/AnalysisView.vue
LTDakin: So when I was searching online I did try to see if there was a way I could convert a UInt8ClampedArray to an image but the method was to put it into a canvas and use toDataUrl() anyways.

However, since I don't need the shared array buffer the analysis view actually doesn't pass it to the worker: `AnalysisView Line 135`, and the worker doesn't create the sharedArray buffer if its not passed: `drawImageWorker.js Line 21`

src/views/AnalysisView.vue
LTDakin: canvas is the only thing passed to create the image and the worker is modified so that it only creates a sharedArrayBuffer if one is sent. The analysis view since it's not merging images doesn't send a sharedArray

src/views/AnalysisView.vue
LTDakin: Oh yeah when testing I found that it wasn't terminating the web workers and they were stacking up so made an issue #161 I can fix it in this PR though

public/drawImageWorker.js
```diff
@@ -1,46 +1,80 @@
...
+  // Re-compute the image and redraw it to the canvas
+  const low16Bit = parseInt(scalePoints[0])
+  const high16Bit = parseInt(scalePoints[1])
+  const scale = 255 / (high16Bit - low16Bit)
+  const len = imageData.data.length
+
+  for (let i = 0; i < len; i++) {
+    const clippedValue = Math.max(low16Bit, Math.min(high16Bit, imageData.data[i]))
+    const normalizedValue = Math.floor((clippedValue - low16Bit) * scale)
+    const gammaCorrected = gammaTable[normalizedValue]
```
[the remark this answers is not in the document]
LTDakin: Yeah this one and `i*4`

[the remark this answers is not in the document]
LTDakin: not sure which one did it since the compiler is a black box but saving look ups and recalculations was the general idea

src/views/AnalysisView.vue
jnation3406: Ah okay I see

src/components/Global/Scaling/HistogramSlider.vue
LTDakin: oops this was incorrect, I thought it was for thumb labels that weren't being used

src/components/Global/Scaling/ImageScaler.vue
LTDakin: I've replaced using the ImageScaler with the histogram slider as it should be and took out the analysis store from the component, created computed properties in the analysis store so its much better now

jnation3406 (approved):
Looks good, thanks for removing the analysis store from the ImageScaler!

PR #170, opened 2025-03-10
fixes polling management for deleted in progress operations

LTDakin:
Bug: Deleting in progress operations didn't stop polling for them so they would keep hitting an error endpoint

Fix: replaced a missing prop with operation Ids that are deleted which is emitted to the parent component who removes the polling for those operations if they exist and refreshes the operation list

### DIFF
--- a/src/components/Global/DeleteOperationDialog.vue
+++ b/src/components/Global/DeleteOperationDialog.vue
@@ -24,7 +24,8 @@ const emit = defineEmits(['update:modelValue', 'itemWasDeleted'])
 const DIALOG_TITLE = 'DELETE OPERATION(S)'
 
 function itemDeleted(){
-  emit('itemWasDeleted', props.operationId)
+  const deletedIds = _.map(props.operations, 'id')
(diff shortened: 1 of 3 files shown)


src/components/Global/DeleteOperationDialog.vue
```diff
@@ -24,7 +24,8 @@ const emit = defineEmits(['update:modelValue', 'itemWasDeleted'])
 const DIALOG_TITLE = 'DELETE OPERATION(S)'
 
 function itemDeleted(){
-  emit('itemWasDeleted', props.operationId)
```
jnation3406, line 27: oops I must have missed this when I made my changes before.

jnation3406 (approved):
Looks good, I don't see anything about the sharedarraybuffer though, is that working for you in staging?

src/components/Global/DeleteOperationDialog.vue
LTDakin: We should add frontend testing when we have time

PR #174, opened 2025-03-11
simplify sws container build

jashan-lco:


LTDakin (approved):
Thanks for improving this

PR #176, opened 2025-03-12
Mobile menu

zemogle:
I noticed that the menu didn't appear when on a small screen.


LTDakin (approved):
Thanks for the fix, header looks good on mobile. I have some work to do to make the rest of the site phone compatible.

PR #181, opened 2025-03-17
adds small_url property check to the image as it is what datalab output small jpgs will be named

LTDakin:
Related to backend changes in https://github.com/LCOGT/ptr_datalab/pull/61


capetillo (approved):
LGTM!

jnation3406 (changes requested):
I believe the `thumbnail_url` is there for datalab generated files? Is that still needed now that we are using `small_url`?

PR #183, opened 2025-03-19
remove old css vars, update to new style sheet vars

LTDakin:
Removed and updated all the old css vars in datalab to the new standard photon ranch colors
Also updated the user account button to look more similar to observe

Project View Page
Screenshot 2025-03-19 at 5 38 01 PM
Screenshot 2025-03-19 at 5 37 52 PM

Data Session Page
Screenshot 2025-03-19 at 5 38 20 PM
Screenshot 2025-03-19 at 5 38 13 PM

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -44,7 +44,7 @@ onMounted(async () => {
 
 <style>
 body {
-  background-color: var(--dark-blue);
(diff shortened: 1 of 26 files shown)


zemogle (approved):
I like the relabeling of the operation buttons too.

PR #184, opened 2025-03-25
Switched small jpeg back to downlaoding original jpeg and not scaled one

jnation3406:


src/views/AnalysisView.vue
```diff
@@ -177,7 +177,7 @@ function updateScaling(min, max){
       <image-download-menu
         :image-name="image.basename"
         :fits-url="image.url || image.fits_url"
-        :jpg-url="analysisStore.imageUrl"
```
jnation3406, line 180: This is the generated scaled jpeg the view was using, not the original large jpeg

LTDakin (approved):
Good catch

PR #187, opened 2025-03-28
Fix/image grid reload on delete

### LINKED ISSUE #185: Deleting operation doesn't refresh the images in the data session
Expected: when I delete an operation the outputs of that operation are removed from my data session

Actual: Images remain until a refresh

LTDakin:
This PR closes #185   where when operations were deleted their corresponding output files were not removed until a refresh. Now in addition to stoping the polling of operations we also remove the images that were the outputs of those operations. 

Other Changes:
- filteredImages is now a computed property so we don't need to call a function to update it when images changes or push outputs to filtered images
- moved `processOperations` to be next to the function that calls it
- Removed lodash from the project since best practice is to now use the js inbuilt array functions
- Moved DeleteSessionDialog to datasession component folder


jnation3406 (approved):
Looks good to me

PR #190, opened 2025-04-03
Feature/output types

LTDakin:
Previously RGB outputs would show up as available input for operations like Medians, Subtractions, and Stacking even though they weren't compatible. In the future datalab might also output other data products like graphs, json, etc.

This PR, paired with a [PR in the backend](https://github.com/LCOGT/ptr_datalab/pull/77) address this by attaching format attributes to datalab outputs and validating input wizard description formats when creating input options.

Instead of 'file', 'fits' is now the format for a valid fits file. RGB outputs have the format of 'image'.

This PR also refactors the code in OperationWizard to follow best practices/designs
- enum page object instead of strings
- groups related functions for engineers
- more descriptive alert for max inputs
- renaming to be more clear what something is, avoid similarly named variables
- operation stored to reduce necessary lookup helpers
- Keeps track of all inputs in one object, instead of images separately that are then added in manually

https://github.com/user-attachments/assets/1d0799d4-be46-4950-8db8-008298b531a4

### DIFF
--- a/src/components/DataSession/OperationWizard.vue
+++ b/src/components/DataSession/OperationWizard.vue
@@ -23,7 +23,7 @@ const availableOperations = ref({})
 const selectedOperation = ref('')
 const selectedOperationInput = ref({})
 const selectedImages = ref({})
-const imagesPerRow = ref(3)
+const IMAGES_PER_ROW = 3
 
 const WIZARD_PAGES = {
   SELECT: 'select',
@@ -288,7 +288,7 @@ function selectOperation(name) {
             <image-grid
               :images="inputDescription.filter ? sortImagesByFilter(inputDescription.filter) : props.images"
               :selected-images="selectedImages[inputKey]"
-              :column-span="calculateColumnSpan(images.length, imagesPerRow)"
+              :column-span="calculateColumnSpan(images.length, IMAGES_PER_ROW)"
               :allow-selection="true"
               @select-image="selectImage(inputKey, $event)"
             />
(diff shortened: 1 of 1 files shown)


src/components/DataSession/OperationWizard.vue
```diff
@@ -257,32 +259,33 @@ function selectImage(inputKey, basename) {
...
           :key="inputKey"
           class="operation-input-wrapper"
         >
           <v-text-field
-            v-if="inputDescription.type !== 'file'"
-            v-model="selectedOperationInput[inputKey]"
+            v-if="inputDescription.type == 'text'"
+            v-model="operationInputs[inputKey]"
             :label="inputDescription.name"
             :type="inputDescription.type"
```
jnation3406, line 274: So this used to work with the type being any html input type, i.e. text, number, date, etc. It's okay to make this change so that no longer works, but we should at least add in a number field for type number since that's a likely one operations will use. I used to have a test operation testing all this but its gone since its not in production...

jnation3406 (approved):
Looks fine, but I would add an option in the wizard for number fields since that is a likely thing that will be used in operations.

src/components/DataSession/OperationWizard.vue
LTDakin: This was kinda vestigial code, since we removed the demo operations we haven't had any operations come up that need numbers or text. I do want to leave it in there though so we can still have the mechanisms for extending it in the future. As it stands nothing will call this. Would you be ok with adding the number field when we get to it? We don't have anything using it right now

jnation3406 (approved):

PR #192, opened 2025-04-09
Feature/rgb stack loading indicators

### LINKED ISSUE #163: Improved loading animations for the RGB stack operation
Right now the RGB stack operation just pops in as the images load, this is pretty janky and could use improvement to better load the image and be less jarring to view.

LTDakin:
Adds loading circles for scaling data to let the user know that scaling is loading. Before it was blank and images would pop in. For larger images this doesn't let the user know that their data is loading and they might think its broken.

Closes #163 

https://github.com/user-attachments/assets/50e1296e-d00c-46fd-8cb7-0f8f63738cae

### DIFF
--- a/src/components/DataSession/OperationWizard.vue
+++ b/src/components/DataSession/OperationWizard.vue
@@ -298,13 +298,12 @@ function selectOperation(name) {
         v-show="page == WIZARD_PAGES.SCALING"
         class="wizard-card"
       >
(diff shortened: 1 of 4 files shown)


jnation3406 (approved):
Looks fine, but should we also hide the default histogram or large composite image with a loader? Either way this is already better than it was before.

PR #195, opened 2025-04-18
Fix/leaflet rawdata unload

### LINKED ISSUE #180: Analysis Performance Issues
Performance issues with the analysis view
- On chrome when the the fetch request to the `raw-data` endpoint returns and the first scaled image is created the image unloads from the leaflet map, pressing the zscale button or adjusting the histogram restores it. Doesn't show up on Firefox or Safari?
- Image sets on the leaflet map seem to be queued. When sliding the histogram quickly you can see the images each take their turn to be placed into the leaflet map until the final state of the slider is reached. It would look nicer if the most recent image was placed into the map instead so you don't have to wait that second to see the state
- A flickering as images are placed into the leaflet map when adjusting the slider on chrome and safari
- The UI is blocked as the first image raw-data is loaded. Figure out what is blocking and either place it into a worker or find a way to unblock the UI as it loads

### LINKED ISSUE #188: RGB Histogram colors wrong
The RGB histogram colors are based on the color of the filter of the input image selected, rather than the color of the channel they represent. For example, if you select a "R" image for all three channels, then all three histograms are red colored. I would expect them to always be red/green/blue colored for the RGB image selection.

LTDakin:
Bug: On the intial load of the raw-data for zscaling the image the image map would go blank. The user would have to adjust the slider or click zscale and then the image would show up.

Cause: There was a race condition with the offscreen canvas and the blob creation for the first run. After it "warmed up" the blob could be created from the canvas. But for the first time the canvas was wrote to it wouldn't finish before the blob was created resulting in a blank image. Since we clean up the previous image with revokeObjectUrl it didn't have a fallback to display,.

Fix: Moved blob creation into the worker so that it always will run after the worker has finished drawing the scaled image to the canvas.

Now the image zscales when the raw-data loads and doesn't create a blank map.

Leaflet map no longer goes blank after zscale data loaded

Closes #188 
Closes #180 

https://github.com/user-attachments/assets/3387bef5-34e4-4af8-8ce1-1e10e3bc9d21

Still working with RGB scaling

Screenshot 2025-04-21 at 11 52 31 AM

### DIFF
--- a/public/drawImageWorker.js
+++ b/public/drawImageWorker.js
@@ -18,6 +18,7 @@ onmessage = function(job) {
     context = canvas.getContext('2d')
     outputImage = new ImageData(payload.width, payload.height)
 
+    // Used for RGB stack, not used in grayscale scaling
     if(payload.sharedArrayBuffer){
       sharedArrayBuffer = payload.sharedArrayBuffer
       sharedArray = new Uint8ClampedArray(sharedArrayBuffer) 
@@ -76,7 +77,7 @@ async function processScalePoints(scalePoints) {
     outputImageData[j + 3] = 255
   }
 
-  context.putImageData(outputImage, 0, 0)
+  await context.putImageData(outputImage, 0, 0)
   const blob = await canvas.convertToBlob()
   postMessage({ blob: blob})
 }
--- a/public/drawImageWorker.js
+++ b/public/drawImageWorker.js
@@ -1,4 +1,4 @@
-var context, imageData, sharedArrayBuffer, sharedArray, gammaTable, outputImage, scalePointMessage
+var canvas, context, imageData, sharedArrayBuffer, sharedArray, gammaTable, outputImage, scalePointMessage
 let hasImageData = false
 let hasCanvas = false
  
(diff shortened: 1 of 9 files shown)


jnation3406 (changes requested):
Changing the worker to post the blob both increases the payload for each run of that worker (the blob is an image that has to be sent between threads), and isn't used in the RawScaledImage use case of this web worker. 

I'm not convinced we need to be generating and sending the blob in the worker to fix the synchronization issue. Maybe its possible to instead hold off on creating the first blob until we have run the scaling worker at least once?

public/drawImageWorker.js
```diff
@@ -71,12 +68,23 @@ function processScalePoints(scalePoints) {
...
+
     outputImageData[j] = gammaCorrected
     outputImageData[j + 1] = gammaCorrected
     outputImageData[j + 2] = gammaCorrected
     outputImageData[j + 3] = 255
   }
-  context.putImageData(outputImage, 0, 0)
-  postMessage({'updateSharedArray': true})
+
+  await context.putImageData(outputImage, 0, 0)
```
jnation3406, line 80: This should also be called only for the non RGB case right?

jnation3406 (changes requested):

public/drawImageWorker.js
jnation3406: wait nvm i guess we are using the canvas there too. But this doesn't return a promise right? So await doesn't mean anything here?

jnation3406 (approved):

PR #206, opened 2025-06-17
Feature/enhanced catalog

### LINKED ISSUE #162: Source catalog user settings and UX
Source catalog just shows 1000 sources currently. But some users may want to keep it at only the top 100, others inside a certain range, and a few might not want to see the source catalog at all.

This feature update should allow for better control of the source catalog tool and also add the settings to be saved with the user profile in the persistent usersettings store so that whatever preferences they have are maintained for all analysis tasks to improve the usability as a tool.

LTDakin:
Enhances the source catalog slider to improve usability

Toggle is now stored in userData settings, so if you want it off/on that will be saved between refreshes
Textfields for flux range now allow you to input exact flux values to define your range at

closes #162 

Screenshot 2025-06-17 at 4 48 35 PM

### DIFF
--- a/src/components/Global/NonLinearSlider.vue
+++ b/src/components/Global/NonLinearSlider.vue
@@ -29,6 +29,11 @@ const props = defineProps({
     type: Function,
     default: (index, stepCount, max) => Math.floor(Math.pow(10, index * Math.log10(max) / stepCount)),
     required: false,
+  },
+  disabled: {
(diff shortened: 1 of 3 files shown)


markBowman (approved):
Shallow review looks good.

capetillo (approved):
Looks good, just one question (out of curiosity)

PR #210, opened 2025-06-20
Feature/additional proposal filters

LTDakin:
Updates to the project view requested by users to make it more useful when looking up images quickly, the submitter query can take some time due to it being in a separate table in the archive so it has a long lookup time to filter

- Adds new filter fields: Target, Submitter
- Fixes a bug where searching an eccentric object would raise data.error in the archive -> clear the ra, dec, and search field without any indication to user
- Update the look of switches 
- fixes the offset table rows

https://github.com/user-attachments/assets/555948a8-3254-411f-86d9-4af882a78d85

### DIFF
--- a/src/views/ProjectView.vue
+++ b/src/views/ProjectView.vue
@@ -221,7 +221,7 @@ watch(() => filters.value.search.value, async () => {
           alertsStore.setAlert('warning', `Simbad ${data.error}`)
         }
         else if(data.eccentricity){
-          alertsStore.setAlert('info', 'LCO archive does not support eccentric object lookup')
+          alertsStore.setAlert('warning', 'LCO archive doesn\'t support Non-Sidereal target lookup')
         }
         else{
           filters.value.ra.value = data.ra_d


src/views/ProjectView.vue
```diff
@@ -163,10 +255,17 @@ onMounted(() => {
 </script>
 
 <template>
-  <div class="proposal-filters">
+  <v-progress-linear
```
LTDakin, line 258: new progress for when filters updated, user knows their query went through

LTDakin (commented):
Explanation comments added

src/directives/vHide.js
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Usage: v-hide="boolean"
+ *
+ * Toggles the 'visibility: hidden' style on an element.
+ * The element remains in the DOM and occupies its space, preventing layout shifts.
+ *
+ * @param {HTMLElement} el - The element the directive is bound to.
+ * @param {object} binding - The binding object, binding.value contains the directive's value.
+ */
+const setVisibility = (el, binding) => {
```
jnation3406, line 12: This is an interesting way of doing this I haven't seen before. Usually I've just set a class re-actively that would have a style set with the visibility on/off, but I guess this breaks it out more explicitly.

src/components/Global/InsetIconSwitch.vue
```diff
@@ -0,0 +1,47 @@
+<script setup>
+import { defineProps, defineEmits } from 'vue'
+
+defineProps({
+  modelValue: {
+    type: Boolean,
+    default: false
+  },
+  trueIcon: {
```
jnation3406, line 9: I feel like rightIcon / leftIcon might better represent this component then true/false?

src/views/ProjectView.vue
```diff
@@ -52,95 +90,149 @@ function selectImage(proposalIndex, basename) {
...
+  // Kicks off a call the the Simbad2k API to get the RA and DEC for the object
+  if(filters.value.search.value){
+    filtersDebounceTimer = setTimeout(async () => {
+      const url = `https://simbad2k.lco.global/${filters.value.search.value}`
+      await fetchApiCall({url: url, method: 'GET', successCallback: (data)=> {
+        if(data.error){
+          alertsStore.setAlert('warning', `Simbad ${data.error}`)
+        }
+        else if(data.eccentricity){
+          alertsStore.setAlert('info', 'LCO archive does not support eccentric object lookup')
```
jnation3406, line 224: This could probably say doesn't support Non-Sidereal target lookup

src/views/ProjectView.vue
```diff
@@ -190,51 +289,24 @@ onMounted(() => {
...
-    />
-    <v-text-field
-      v-model="dec"
-      label="DEC"
-      clearable
+      v-for="filter in Object.values(filters).filter(f => f.label && f.type !='hidden')"
+      :key="filter.label"
+      v-model="filter.value"
+      :class="filter.class || ''"
+      :label="filter.label"
```
jnation3406, line 296: Nice that simplifies things a bit!

jnation3406 (approved):
Looks good, just a few naming comments.

src/components/Global/InsetIconSwitch.vue
LTDakin: I named it that just to keep it consistent with vuetify's v-switch props, I think for this case I'll keep it the same
```
<v-switch>
    :false-icon="falseIcon"
    :true-icon="trueIcon"
</v-switch>

PR #224, opened 2025-07-23
Assortment of UI changes

LTDakin:
A hodge podge of tweaks to improve the Datalab UI

- Change Submitter filter field to `Submitter User ID
- Change default time range in project view from 5 days to 1 month
- Reorder project list view to img, target, filter, time, site, basename
- Resize source circles in Analysis view to be smaller
- Replace v-html prop with slots to remove XSS attack warning

### DIFF
--- a/src/components/Project/ImageList.vue
+++ b/src/components/Project/ImageList.vue
@@ -23,7 +23,7 @@ const emit = defineEmits(['selectImage'])
 const headers = ref([
   { title: 'IMAGE', align: 'start', sortable: false, key: 'url' },
   { title: 'TARGET', align: 'start', sortable: true, key: 'target_name' },
-  { title: 'FILTER', align: 'start', sortable: true, key: 'FILTER' },
+  { title: 'FILTER', align: 'start', sortable: true, key: 'primary_optical_element' },
   { title: 'TIME', align: 'start', sortable: true, key: 'observation_date' },
   { title: 'SITE', align: 'start', sortable: true, key: 'site_id' },
   { title: 'IMAGE NAME', align: 'start', sortable: true, key: 'basename' },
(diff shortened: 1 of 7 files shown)


src/components/Project/ImageList.vue
```diff
@@ -22,12 +22,11 @@ const emit = defineEmits(['selectImage'])
 
 const headers = ref([
   { title: 'IMAGE', align: 'start', sortable: false, key: 'url' },
-  { title: 'FILTER', align: 'start', sortable: true, key: 'FILTER' },
-  { title: 'IMAGE NAME', align: 'start', sortable: true, key: 'basename' },
   { title: 'TARGET', align: 'start', sortable: true, key: 'target_name' },
+  { title: 'FILTER', align: 'start', sortable: true, key: 'FILTER' },
```
jnation3406, line 26: use `primary_optical_element` instead of `FILTER` for getting the filter from the /frames query

src/views/ProjectView.vue
```diff
@@ -63,11 +63,12 @@ const filters = ref({
   },
   submitter: {
     value: route.query.submitter || '',
-    label: 'Submitter',
+    label: 'Submitter User ID',
   },
 })
 let filtersDebounceTimer
 const FILTER_DEBOUNCE = 600
+const DEFAULT_DATE_RANGE_DAYS = -30
```
jnation3406, line 71: Seems strange having a negative value for a constant that says its days. Maybe just make this positive but put a negative sign in front of it when you want to use it to subtract days from something?

jnation3406 (approved):
Looks good, didn't realize the slot thing could be used instead of passing the html through but that is much cleaner. Just a few small comments.

PR #225, opened 2025-07-30
Feature/grid cards

### LINKED ISSUE #216: Image Grid Info/Actions below image
As a user it's unintuitive to know that I need to click on an image to open it in analysis view for things like downloading. I would like to see in image grid a little blurb of info below the image that tells me about it, and lets me easily download or open the analysis view.

Tasks
- [ ] Add a "polaroid" type card around images
- [ ] Add Filter and name to the card
- [ ] Add download and analysis icon buttons to the card

LTDakin:
Adds "Cards" to the image grid that have a couple of things
- target name of the image
- an analysis icon button `mdi-eye`
- a download button that supports `fits` and `jpg` downloads

Works for the `Datasession` view as well and datalab outputs

Screenshot 2025-07-30 at 2 17 05 PM

Closes #216


LTDakin:
You can still download a TIF from the analysis view, I looked into adding it to the extra button but it would take some reorganizing that is out of the scope for this story, could make it a new story to add tif downloads to the card as well

src/components/Global/ImageGrid.vue
```diff
@@ -116,15 +108,35 @@ watch(() => props.images, () => {
...
-        />
-      </v-col>
-    </template>
-    <template v-else>
+        <div
+          v-if="props.enableImageCards"
+          class="d-flex flex-row ga-2 align-center mt-2"
+        >
+          <p class="text-subtitle-2 mr-auto prevent-select single-line-text">
+            {{ image.target_name }}
```
jnation3406, line 116: Datalab output images don't have a target_name field (or a filter field for that matter). Maybe you can put something else here? Maybe what operation it came from (i.e. Median, Subtraction, RGB) or something else relevant.

jnation3406 (approved):
Looks good to me, just think maybe there is something we can do for Datalab output images too.

PR #229, opened 2025-08-12
Add drag and drop for image selection in the operation wizard

jnation3406:
First cut at adding drag and drop support - haven't tested this on Ipad yet, just on my computer. I refactored the image stuff out of ImageGrid.vue into ThumbnailImage.vue to reuse that portion with the new MultiImageInputSelector.vue component

[datalab_image_selection_dnd.webm](https://github.com/user-attachments/assets/a472bdc5-b9fe-4331-8ffb-dfe71bac2197)

### DIFF
--- a/src/components/DataSession/MultiImageInputSelector.vue
+++ b/src/components/DataSession/MultiImageInputSelector.vue
@@ -1,0 +1,190 @@
+<script>
+import { useThumbnailsStore } from '@/stores/thumbnails'
+import { useConfigurationStore } from '@/stores/configuration'
+import ThumbnailImage from '@/components/Global/ThumbnailImage.vue'
+import {Drag,DropList} from 'vue-easy-dnd'
+import { ref } from 'vue'
+
+export default {
+  name: 'MultiImageInputSelector',
+  components: {
+    Drag, DropList, ThumbnailImage
+  },
+  props: {
+    inputDescriptions: {
+      type: Object,
+      default: () => {}
+    },
+    selectedImages: {
+      type: Object,
+      default: () => {}
+    },
+    images: {
+      type: Array,
+      default: () => []
+    }
+  },
+  emits: ['insertSelectedImage', 'removeSelectedImage'],
+  data: function () {
+    return {
+      thumbnailsStore: useThumbnailsStore(),
+      configurationStore: useConfigurationStore(),
+      imageDetails: {},
+      filterToColor: {
+        'r': 'red',
+        'rp': 'red',
+        'ip': 'red',
+        'h-alpha': 'purple',
+        'v': 'green',
+        'gp': 'green',
+        'oiii': 'green',
+        'b': 'blue',
+        'sii': 'blue'
+      }
+    }
+  },
+  computed: {
+    columnSize: function() {
+      const inputCount = Object.keys(this.inputDescriptions).length
+      return Math.floor(12 / inputCount)
+    }
+  },
+  methods: {
+    insert(inputKey, event) {
+      if (inputKey !== 'all' && ! this.selectedImages[inputKey].includes(event.data)) {
+        this.$emit('insertSelectedImage', inputKey, event.data)
+      }
+    },
+    remove(inputKey, value) {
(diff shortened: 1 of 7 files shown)


src/components/DataSession/MultiImageInputSelector.vue
```diff
@@ -0,0 +1,228 @@
+<script>
+import { useThumbnailsStore } from '@/stores/thumbnails'
+import { useConfigurationStore } from '@/stores/configuration'
+import ThumbnailImage from '@/components/Global/ThumbnailImage.vue'
+import {Drag,DropList} from 'vue-easy-dnd'
+import { ref } from 'vue'
+
+export default {
```
LTDakin, line 8: This is using the options api which is fine but for consistency we use the composition api in the rest of the application so we should stick to that in the future

src/components/DataSession/MultiImageInputSelector.vue
```diff
@@ -0,0 +1,228 @@
...
+      default: () => []
+    }
+  },
+  emits: ['insertSelectedImage', 'removeSelectedImage'],
+  data: function () {
+    return {
+      thumbnailsStore: useThumbnailsStore(),
+      configurationStore: useConfigurationStore(),
+      imageDetails: {},
+      filterToColor: {
```
LTDakin, line 33: we have a filterToColor util in /utils/common.js that should be used instead

src/components/DataSession/MultiImageInputSelector.vue
```diff
@@ -0,0 +1,228 @@
...
+        <v-card-text>
+          <drop-list
+            :items="images"
+            mode="cut"
+            :row="true"
+            :no-animations="true"
+            @insert="insert('all', $event)"
+          >
+            <template #item="{item}">
+              <drag
```
LTDakin, line 164: At least for me on Mac/Chrome I can't see the dragged image, but it does work to drag, this is fine but could we also keep the click functionality? For multiple filters like RGB it could loop through each next filter when you click. But if thats more involved this could be another story to bring click functionality back

LTDakin (changes requested):
The refactor and just overall improvement to the selection logic is really nice. 
Would like to see the click logic still stay around if possible as an alternative to drag.

Also we have a eslintrc.js file in the project for linting, you can add this [eslint vscode extention](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) so the linting is automatic

src/components/DataSession/MultiImageInputSelector.vue
jnation3406: Yeah, I prefer the composition api but this vue-easy-dnd says it only works with the options api because it has mixins or something

src/components/DataSession/MultiImageInputSelector.vue
jnation3406: Yeah so you don't see the image while dragging until it is over a drop surface. Once its over a drop surface but before you drop it, you can see it in the drop surface. I can see if there is a way to make it visible during drag but nothing popped out in the API, its a pretty bare-bones api. 

For click functionality, it seems clunky for the multiple filter case since you would need to somehow click on the filter slot first to select that slot, then click on an image to place it in that slot. Thus requiring two clicks where it used to require one, but I can't see how to avoid that since we removed the separate set of images for each filter slot... I don't really understand what you mean by looping through each filter when you click but maybe you can describe it to me in the next meeting.

jnation3406: Okay i figured out that showing the image during dragging was supposed to happen by default, but was being blocked by the z-index of the wizard dialog. I checked the source and found the drag-image had a z-index of 1000, so I moved the z-index of the wizard dialog to 999 from 2400. Its a bit arbitrary so should be fine.

LTDakin (approved):
Ok with the drag I'm much happier leaving the click out, I think I may come back to it though and implement it another time. Last thing is that I can see the color images as available input for the operations, they should be excluded since we can't use them and it will cause an error.

PR #107, opened 2024-09-25
Project View overhaul/Refactor

LTDakin:
Problem: The project page had some vestigial design choices from when we first were scaffolding data lab. This PR redoes the project view and refactors a lot of the old code into a design that serves users (recommended by Edward).

Theres a lot of changes so I'll be listing them by file, but theres also countless decisions and changes to improve the user experience which would be too many to list out so I encourage you to clone the branch and test it out for yourself! 

Screenshot 2024-09-25 at 4 19 28 PM
Screenshot 2024-09-25 at 4 19 39 PM
Screenshot 2024-09-25 at 4 20 24 PM

packages
- updated vuetify to use 3.6.0 which also updates package-json

Image Grid
- fixed double click, now works with a debouncing timer to open image analysis
- conditional rendering of grid based on prop value, renders a loading skeleton if false
- Add alert for failed analysis

User Menu
- move grid/table toggle to project view filters
- add color to v-switch

Create Session Dialog
- Previously inside the project view component, now lives in its own
- refactored code
- updated UI includes last modified info, guiding text, closure top right

Image List
- Changed select logic to work with the image grid index pattern
- added fields: SITE, TELESCOPE, TARGET
- reordered fields to put most important info to the left
- Added analysis component to image list, activated when image clicked
- Conditional rendering of table if data, otherwise displays a table skeleton loader

Project Bar (DELETED)
- Took up too much space, replaced with expansion panels

Project Selector (DELETED)
- Took up too much space, replaced with expansion panels

Settings Store (DELETED)
- Made sense to remove as it had a mix of values, most of which where only used by the ProjectView, and others belonged to different stores

userData
- Now holds rest of settings that are stored for user
- openProposals remembers which proposals the user looked at recently and keeps it open for them
- mostRecentSessionID moved from Settings store
- refactored auth method to be shorter

api
- changed failCallback to use handleError as default method

DataSessionView
- changed settings store usage to use different store

ProjectView
- Projects are now called Proposals to stick with LCO naming
- Proposals are contained in expansion panels
- Carousel replaced with imageGrid
- switch from grid to table is now on project view
- Refactored functions and variables
- Filtering by Start Date, End Date, search by point, SImbad Search that autofills point search
- Ra and Dec debounces so user has time to type in
- Create Session Dialog logic pulled out

PR #236, opened 2025-09-12
decouple light curve from periodogram so it still works if period fails

LTDakin:
This leads to more graceful failure, program won't load endlessly anymore and light curve for the star will still be shown.

- Decouples the `magTimeSeries` data from the periodogram so we can still display light curves when the period fails to calculate
- Checks that there is period data before doing the period folding work

Paired with: https://github.com/LCOGT/ptr_datalab/pull/107


jnation3406 (approved):
I think this is a good first step but it would be better to know that the peroidogram failed since the user is explicitly asking for it

PR #244, opened 2025-12-02
Feature/periodogram improvements

jnation3406:
There are a bunch of changes in here so it might be easiest to just talk it over, but I'll try to at least put an outline here:

- Support for other input types in `OperationWizard` - including new 'source' type with its own `SourceInputWidget` which can do a source lookup by name and provides a name/ra/dec to the operation
- Changes to `MultiImageInputSelector` used by the wizard to allow restricting by filter, and allow selecting all images of a given filter from a dropdown - keyed on new keywords in the operations wizard description
- Refactor `ImageGrid` into a more generic `OperationOutputGrid` that can support both images and non-image data output in a grid view. The `ImageGrid` still exists since it is used for input image selection, which is kind of different now from showing output
- New `DataOutput` which encapsulates generic data output (as opposed to image output) from operations. It has code to show a sparkline plot of light curve or period data as well as the source / operation name
- Rename `AnalysisView` into `ImageAnalysisView` and only launch it from image data.
- Create `DataAnalysisView` which is launched from `DataOutput` data. This is where your code for showing the light curve, periodogram, and phased light curve will go. The more generic and piecemeal this is the better, since we will likely have to extend this to show other data output stuff later with new operations.


src/components/DataSession/Operation/MultiImageInputSelector.vue
```diff
@@ -54,14 +55,54 @@ const fitsImages = computed(() => {
...
+  if (props.inputDescriptions[inputKey].filter_options) {
+    return filterOptions.filter(item => props.inputDescriptions[inputKey].filter_options.includes(item))
+  }
+  else{
+    return filterOptions
+  }
+}
+
 // Image dragged into the selected images area
 function insert(inputKey, index, event) {
```
capetillo: just a suggestion, would it make sense to make this function into a utils function since other operations use `insert` logic?

capetillo (approved):
LGTM! P.S. Have we figured out why the dragging scale lost its blue color?

PR #247, opened 2026-01-07
Feature/period and dpc

capetillo:
## FEATURE: Dynamic phased light curve based on periodogram selection & download of charts

**Background:**
While the light curve was generated based on the highest peak of the periodogram, it's useful for users to select anywhere on the chart and have a light curve be generated based on their selection. 

**Implementations**
Moved logic around, particularly logic from the Analysis store because this logic no longer belongs in the Analysis view. A lot of files updated was code removed. 
When user clicks on a point on the line of the periodogram, the listener emits a value to the phased light curve component and it regenerates the chart based on the values passed. 

**UPDATE NOT CAPTURED ON VISUALS**
Flipped values on magnitude in phased light curve chart

### VISUALS
**Screen recording of flow**

https://github.com/user-attachments/assets/111b59f2-846d-4327-9576-16e1eed9caa0

**Downloaded charts (as pngs)**
image

image

image

### DIFF
--- a/src/components/Analysis/VariableStarDialog.vue
+++ b/src/components/Analysis/VariableStarDialog.vue
@@ -62,7 +62,6 @@ function dispatchVariableAnalysis() {
     }),
     target_coords: props.coords, // the target coordinates
   })
-  analysisStore.variableStarData.loading = true
   emit('closeDialog')
 }
 


src/views/DataAnalysisView.vue
```diff
@@ -20,6 +116,10 @@ const title = computed(() => {
   return text
 })
 
+const computedProps = computed(() => {
```
jnation3406: This doesn't look good to me... props are already reactive so why are you making them computed?

src/views/DataAnalysisView.vue
```diff
@@ -39,12 +139,44 @@ const title = computed(() => {
...
+            @period-selected="handlePeriodSelected"
+          />
+        </v-row>
+        <v-row class="analysis-row">
+          <period-plot
+            :variable-star-data="variableStarData"
+            :periodogram-data="periodogramData"
+            class="period-plot"
+          />
+        </v-row>
```
jnation3406: I think we should fill the space better. Maybe having a larger light curve on the left column, then a stacked periodogram / phased light curve on the right? Or a stacked light curve / periodogram on the left and larger phased light curve on the right? Or maybe a larger periodogram since that is the one the user is interacting with?

Either way, just think we can fill up the space a bit better, which might be important for a tablet to see all 3 plots on screen at once without scrolling down.

src/stores/analysis.js
```diff
@@ -35,6 +35,15 @@ export const useAnalysisStore = defineStore('analysis', {
       fluxFallback: false, // flag to indicate if flux fallback is used
       excludedImages: [], // list of excluded images for the variable star
     },
+    periodogram: {
```
jnation3406: Is this old code? Also, I think we can remove a bunch of other stuff related to the variableStarData in here, right?

src/components/Analysis/PeriodPlot.vue
```diff
@@ -1,9 +1,19 @@
 <script setup>
 import Chart from 'chart.js/auto'
-import { ref, watch, computed } from 'vue'
-import { useAnalysisStore } from '@/stores/analysis'
+import { ref, watch, computed, defineProps, onMounted } from 'vue'
```
jnation3406: Should this be called a PhasedLightCurvePlot instead of PeriodPlot?

src/components/Analysis/PeriodogramPlot.vue
```diff
@@ -0,0 +1,230 @@
...
+        class="download-btn"
+        title="Download as PNG"
+        @click="downloadChartAsPNG(chart, 'periodogram-plot.png', 'Periodogram')"
+      />
+    </div>
+  </div>
+</template>
+
+<style scoped>
+.wrapper {
```
jnation3406: All 3 of these *Plot.vue files share similar styling I think? So that's a sign that they could potentially re-use code, whether that be a re-used full component, or just a re-used stylesheet they import.

src/views/DataAnalysisView.vue
capetillo: there's no scrolling, but that just means that the graphs are small. I'll work on something

src/components/Analysis/LightCurvePlot.vue
```diff
@@ -128,10 +138,50 @@ function createChart() {
...
-    class="light-curve-plot"
-  />
+  <div class="wrapper">
+    <h1 class="title-lc">Light Curve</h1>
+    <div class="light-curve-plot-wrapper">
+      <canvas
+        ref="lightCurveCanvas"
+        class="light-curve-plot"
+      />
+      <v-btn
```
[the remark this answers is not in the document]
capetillo: alright, eaaaasy fix

src/components/Analysis/PeriodPlot.vue
capetillo: yep

src/views/DataAnalysisView.vue
```diff
@@ -12,6 +15,99 @@ const props = defineProps({
...
+      fluxFallback: data.flux_fallback,
+      excludedImages: data.excluded_images,
+      magnitudeTimeSeries: magnitudeTimeSeries,
+    }
+  }
+
+  // Pairing frequency and power data
+  if (frequency && power && frequency.length === power.length) {
+    const pairs = frequency.map((f, i) => ({ f: Number(f), p: Number(power[i]) }))
+    // Sorting by frequency
```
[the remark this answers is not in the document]
capetillo: nope, you're right

src/stores/analysis.js
```diff
@@ -34,7 +34,7 @@ export const useAnalysisStore = defineStore('analysis', {
       falseAlarmProbability: 0, // false alarm probability for the variable star
```
jnation3406: I think some of these variables relating to the variable star stuff should be no longer used and can be removed too.

PR #252, opened 2026-02-17
Adds dropdown option to request reprojection for stacked images

capetillo:
Adds dropdown menu with the options of `reprojection` and `simple` for stacked data operation

image


src/components/DataSession/Operation/OperationWizard.vue
```diff
@@ -371,13 +384,13 @@ function updateScaling(channelIndex, zmin, zmax) {
         >
           <v-col
             v-for="(inputDescription, inputKey) in group"
-            class="pb-0"
             :key="'input-col-' + inputKey"
+            class="pb-0"
```
jnation3406: perhaps we should add a cols=6 for this v-col since this is meant for two columns per row for inputs

jnation3406 (approved):

PR #255, opened 2026-04-20
Feature/slideshow

capetillo:
## Feature: Slideshow

**Background:**
Rachel would like to see the movement of moons throughout a period of time. Though she would like in the future for this to be an operation where users select a number of images and the output is a video/gif/animation of them going one after the other. A happy medium in the meantime is to add previous and next buttons and sort images chronologically.

**Implementation:**
First, in Data Session View, images are now sorted chronologically. This makes the next and previous button work as intended.
There are two modes that users can view their images: Analysis mode and View mode. In analysis mode, things remain the same. The previous and next buttons are not available here because it takes a really long time to load. In view mode, the user can go back and forth and then switch back into Analysis mode to analyze the image they are currently viewing. 

### VISUALS
**Analysis mode**
image

**View mode**
image

**Screen recording**

https://github.com/user-attachments/assets/65557627-676c-404d-9205-dd896dd9ec99

UPDATED SCREEN RECORDING

https://github.com/user-attachments/assets/1ee538d8-7a56-46b0-b5ca-7981f1e0352b

### DIFF
--- a/src/components/Analysis/FitsHeaderTable.vue
+++ b/src/components/Analysis/FitsHeaderTable.vue
@@ -13,12 +13,20 @@ const tableHeaders = [
   { title: 'Value', sortable: false, key:'1' },
 ]
 
+const basenameSequence = computed(() => {
+  const basename = analysisStore.image?.basename || ''
+  const match = basename.match(/^[^-]+-[^-]+-[^-]+-([^-]+)-e91$/)
+
+  return match?.[1] || 'Unknown'
+})
+
 // Loopable chip dict for v-chips
 const headerChips = computed(() => [
   { icon: 'mdi-earth', text: siteIDToName(analysisStore.headerData.SITEID) },
   { icon: 'mdi-telescope', text: analysisStore.headerData.TELID },
   { icon: 'mdi-camera', text: analysisStore.headerData.INSTRUME },
-  { icon: 'mdi-clock', text: new Date(analysisStore.headerData.DATE).toLocaleString() }
+  { icon: 'mdi-clock', text: new Date(analysisStore.headerData.DATE).toLocaleString() },
+  { icon: 'mdi-numeric', text: basenameSequence.value }
 ])
 
 </script>
--- a/src/components/Analysis/FitsHeaderTable.vue
+++ b/src/components/Analysis/FitsHeaderTable.vue
@@ -1,6 +1,6 @@
 <script setup>
 import { ref, computed } from 'vue'
-import { siteIDToName } from '@/utils/common'
+import { basenameToSequence, siteIDToName } from '@/utils/common'
 import { useAnalysisStore } from '@/stores/analysis'
 
 const analysisStore = useAnalysisStore()
@@ -13,12 +13,7 @@ const tableHeaders = [
   { title: 'Value', sortable: false, key:'1' },
 ]
 
-const basenameSequence = computed(() => {
-  const basename = analysisStore.image?.basename || ''
-  const match = basename.match(/^[^-]+-[^-]+-[^-]+-([^-]+)-e91$/)
-
-  return match?.[1] || 'Unknown'
-})
+const basenameSequence = computed(() => basenameToSequence(analysisStore.image?.basename || ''))
 
 // Loopable chip dict for v-chips
 const headerChips = computed(() => [
--- a/src/components/Analysis/ViewMode.vue
+++ b/src/components/Analysis/ViewMode.vue
@@ -1,0 +1,113 @@
+<script setup>
+import { ref, watch } from 'vue'
+
+const emit = defineEmits(['updateCurrentIndex'])
+
+const props = defineProps({
+  currentIndex: {
+    type: Number,
+    required: true
+  },
+  images: {
+    type: Array,
+    required: true
+  }
+})
+
+const imageUrl = ref(null)
+const loading = ref(false)
+
+async function fetchThumbnail(basename) {
+  loading.value = true
+  try {
+    const response = await fetch(`https://archive-api.lco.global/thumbnails/?frame_basename=${basename}&size=large`)
+    const json = await response.json()
+    imageUrl.value = json.results?.[0]?.url || null
+  } catch (e) {
+    imageUrl.value = null
+  } finally {
+    loading.value = false
+  }
+}
+
+watch(() => props.currentIndex, (newIndex) => {
+  const basename = props.images[newIndex]?.basename
+  if (basename) fetchThumbnail(basename)
+  else imageUrl.value = null
+}, { immediate: true })
+
+function showPrev() {
+  if (props.currentIndex > 0) {
+    emit('updateCurrentIndex', props.currentIndex - 1)
+  }
+}
+function showNext() {
+  if (props.currentIndex < props.images.length - 1) {
+    emit('updateCurrentIndex', props.currentIndex + 1)
+  }
+}
+</script>
+
+<template>
+  <div class="viewmode-container">
+    <v-btn
+      icon="mdi-chevron-left"
+      :disabled="props.currentIndex === 0"
+      @click="showPrev"
+    />
+    <div class="image-wrapper">
+      <v-progress-circular
+        v-if="loading"
+        indeterminate
+        color="primary"
+      />
+      <img
+        v-else-if="imageUrl"
+        crossorigin="anonymous"
+        :src="imageUrl"
+        alt="Thumbnail"
+        class="thumbnail-img"
+      >
+      <div
+        v-else
+        class="no-image"
+      >
+        No image available
+      </div>
+    </div>
+    <v-btn
+      icon="mdi-chevron-right"
+      :disabled="props.currentIndex === props.images.length - 1"
+      @click="showNext"
+    />
+  </div>
+</template>
+
+<style scoped>
+.viewmode-container {
+  display: flex;
+  align-items: center;
+  justify-content: flex-start;
+  flex: 1;
+  width: 100%;
+  height: 100%;
+}
+.image-wrapper {
+  flex: 1;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  min-height: 400px;
+}
+.thumbnail-img {
+  max-width: 100%;
+  max-height: 80vh;
+  border-radius: 8px;
+  box-shadow: 0 2px 8px rgba(0,0,0,0.15);
+}
+.no-image {
+  color: var(--text);
+  font-size: 1.2rem;
+  text-align: center;
+}
+</style>
--- a/src/components/Analysis/ViewMode.vue
+++ b/src/components/Analysis/ViewMode.vue
@@ -9,14 +9,6 @@ const props = defineProps({
   image: {
     type: Object,
     required: true
-  },
-  hasPrevious: {
-    type: Boolean,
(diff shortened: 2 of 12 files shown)


src/components/Analysis/ViewMode.vue
```diff
@@ -0,0 +1,113 @@
...
+  images: {
+    type: Array,
+    required: true
+  }
+})
+
+const imageUrl = ref(null)
+const loading = ref(false)
+
+async function fetchThumbnail(basename) {
```
jnation3406: We want to use the image caching mechanisms here. Its possible we already have the large thumbnail cached... Check out the `stores/thumbnail.js` `cacheImage` which will either get/cache and return the image or simply return the cached image if its already in the cache.

src/components/Analysis/ViewMode.vue
```diff
@@ -0,0 +1,113 @@
...
+  else imageUrl.value = null
+}, { immediate: true })
+
+function showPrev() {
+  if (props.currentIndex > 0) {
+    emit('updateCurrentIndex', props.currentIndex - 1)
+  }
+}
+function showNext() {
+  if (props.currentIndex < props.images.length - 1) {
```
jnation3406: For toggling back and forth between a few images, you really don't want to click but just use the arrow keys. So I think adding a listener on this page for keyboard arrow left/right and triggering the show* function would be a big help.

src/components/Analysis/FitsHeaderTable.vue
```diff
@@ -13,12 +13,20 @@ const tableHeaders = [
   { title: 'Value', sortable: false, key:'1' },
 ]
 
+const basenameSequence = computed(() => {
```
jnation3406: You call this in multiple places so it should be refactored into a utils js

src/views/ImageAnalysisView.vue
```diff
@@ -145,39 +278,80 @@ async function instantiateScalerWorker(){
...
+  if (val === 'Analysis Mode') {
+    const image = resolveImageByBasename(selectedBasename.value) || props.image
+    await loadActiveImage(image)
   }
 }
 
 </script>
 <template>
+  <div class="mode-dropdown-center">
+    <select
```
jnation3406: Since there are 2 options, I think this should be a toggle button instead of a select menu

src/components/DataSession/DataSession.vue
```diff
@@ -281,6 +287,7 @@ watch(
         </v-col>
         <operation-output-grid
           :operation-outputs="filteredImages"
+          :images="sortByObservationDate(props.data.input_data)"
```
jnation3406: I hate this - we basically have two props for different filters on the same set of images!

jnation3406 (changes requested):
A few usability comments, but then more generally I don't like how we have to pass in the entire list of images to a chain of components that want to show you a single image. What do you think about instead having the ImageAnalysisView emit a next/previous event, which the parent (OperationOutputGrid or ImageGrid or ImageList) listen to to then update the `image` prop for their child ImageAnalysisView? I think that will be much less confusing and help us when debugging down the road, since the analysis view will still operate on one image at a time, and that image will be known and updated as you toggle.

src/views/ImageAnalysisView.vue
```diff
@@ -52,33 +70,144 @@ const filteredCatalog = computed(() => {
...
+  return image.largeCachedUrl
+}
+
+function resolveImageByBasename(basename) {
+  if (!basename) return null
+
+  return props.images.find((image) => image.basename === basename) || null
+}
+
+async function loadActiveImage(image) {
```
jnation3406, line 148: This function and the one above and all the "activeImage" stuff would go away if we have the parent of this handle changing the prop "image" on next/prev instead of doing it in the child.

src/components/Analysis/ViewMode.vue
capetillo: hmm i see, yeah i agree

src/views/ImageAnalysisView.vue
capetillo: yes!

src/components/Analysis/ViewMode.vue
capetillo: okay, I think we need both because the ipad won't necessarily have arrows. I will add the event listener for arrows as well as swiping but I think leaving the buttons is also necessary

jnation3406: Yeah I have no issues with the buttons, just wanted the keyboard controls in addition!

src/components/Analysis/ViewMode.vue
```diff
@@ -1,72 +1,150 @@
...
 }
 function showNext() {
-  if (props.currentIndex < props.images.length - 1) {
-    emit('updateCurrentIndex', props.currentIndex + 1)
+  if (props.hasNext) {
+    emit('next')
+  }
+}
+
+function handleTouchStart(event) {
```
jnation3406, line 66: What is this touch stuff? I though touching the arrow buttons already worked? Is this for keyboard arrow key controls?

src/components/Analysis/ViewMode.vue
```diff
@@ -1,72 +1,150 @@
...
       />
       <img
-        v-else-if="imageUrl"
+        v-if="imageUrl"
         crossorigin="anonymous"
         :src="imageUrl"
         alt="Thumbnail"
         class="thumbnail-img"
+        :class="{ 'thumbnail-img--loading': loading }"
+        @load="handleImageLoad"
```
jnation3406: these names might be a little more clear as like `doneLoadingImage` `errorLoadingImage` or something?

src/components/DataSession/DataSession.vue
```diff
@@ -24,8 +24,14 @@ const props = defineProps({
...
 
+function sortByObservationDate(items) {
+  return [...items].sort((a, b) => {
+    return new Date(a.observation_date) - new Date(b.observation_date)
+  })
+}
+
 const operations = ref([...props.data.operations])
-const items = ref([...props.data.input_data])
+const items = ref(sortByObservationDate(props.data.input_data))
```
jnation3406, line 34: The items wont be reactive to changes in the props... I think this is okay? But we should test by loading into your data session, than going to project view and adding more input files to your data session, then back to your datasession and make sure those new files show up.

src/components/Global/ImageGrid.vue
```diff
@@ -37,25 +37,56 @@ const showAnalysisDialog = ref(false)
 const imageDetails = ref({})
 const analysisImage = ref({})
 
-async function ensureLargeCachedUrl(image) {
-  if (!image.largeCachedUrl) {
-    const url = image.large_url || image.largeThumbUrl || ''
-    image.largeCachedUrl = await thumbnailsStore.cacheImage('large', configurationStore.archiveType, url, image.basename)
+function getAnalysisImage(image) {
```
jnation3406: This function exists to just add 2 properties to the image itself? But its only called on demand to load that image into the analysis view. I would prefer we just add these properties if you want them added.

Actually better yet, maybe we shouldn't care about bounds first/last and just let you scroll infinitely through the images?

src/components/Global/OperationOutputGrid.vue
```diff
@@ -37,19 +33,42 @@ const imageDetails = ref({})
...
 
-async function ensureLargeCachedUrl(image) {
-  if (!image.largeCachedUrl) {
-    const url = image.url || image.large_url || image.largeThumbUrl || ''
-    image.largeCachedUrl = await thumbnailsStore.cacheImage('large', configurationStore.archiveType, url, image.basename)
+const imageOperationOutputs = computed(() => {
+  return props.operationOutputs.filter((operationOutput) => isImage(operationOutput))
+})
+
+function getAnalysisImage(image) {
```
jnation3406: Same comment as above, I think this method would go away if we allow infinite scrolling through images

src/components/Global/OperationOutputGrid.vue
```diff
@@ -37,19 +33,42 @@ const imageDetails = ref({})
...
+}
+
+function imageArchiveSource(image) {
+  if (image?.source && image.source !== 'archive') {
+    return image.source
   }
+  return configurationStore.archiveType
+}
+
+async function ensureLargeCachedUrl(image) {
```
jnation3406: These methods are the same between ImageGrid and this so could be reused

src/components/Project/ImageList.vue
```diff
@@ -34,6 +34,22 @@ const alertsStore = useAlertsStore()
 const thumbnailsStore = useThumbnailsStore()
 const configurationStore = useConfigurationStore()
 
+function getAnalysisImage(image) {
+  const currentIndex = props.images.findIndex((candidate) => candidate.basename === image?.basename)
+  return {
+    ...image,
+    hasPrevious: currentIndex > 0,
+    hasNext: currentIndex > -1 && currentIndex < props.images.length - 1,
+  }
```
jnation3406: same comments as the ImageGrid and OperationOutputGrid

src/components/Analysis/ViewMode.vue
jnation3406: okay I see the keyboard stuff in the ImageAnalysisView

jnation3406 (changes requested):
I think this is almost there, just a comment on reusing some of the code between the ImageGrid/ImageList/OperationOutputGrid for view mode stuff, and I think we could simplify some more code if we just allow endless scrolling left or right looping through the images instead of bounding it and stopping you.

src/components/Analysis/ViewMode.vue
capetillo: it's swiping for tablet (codex recommended this)

src/components/Global/ImageGrid.vue
capetillo: oh okay!

src/components/DataSession/DataSession.vue
capetillo: https://github.com/user-attachments/assets/4b620439-d537-4ab1-b5e5-35307c9be013

src/components/Global/ImageGrid.vue
```diff
@@ -37,54 +38,28 @@ const showAnalysisDialog = ref(false)
...
 }
 
 async function showAdjacentImage(direction) {
+  if (!props.images.length) return
+
   const currentIndex = props.images.findIndex((image) => image.basename === analysisImage.value?.basename)
   if (currentIndex < 0) return
 
-  const nextIndex = currentIndex + direction
+  const nextIndex = (currentIndex + direction + props.images.length) % props.images.length
```
jnation3406, line 58: I think you don't need the ` + props.images.length` inside the first paranthesis

jnation3406 (approved):
Looks great, I love seeing code removed!! Just one comment that doesn't change functionality.

src/components/Global/ImageGrid.vue
capetillo: no, we do need it. If we have 5 images and we're on index `0` and were to move backwards `(0 - 1) % 5 = -1` and there's no -1 index

capetillo: with the `+ props.images.length` and with the same example as above we get `(0 - 1 + 5) % 5 = 4`

PR #256, opened 2026-04-23
Feature/centroiding

capetillo:
##FEATURE: Centroiding

**Background:**
DataLab needs centroiding tools in the Analysis view to obtain the centroids of objects that we cannot get from BANZAI. 

**Implementation**
The centroiding is treated as a leaflet map tool. When the tool is active, the user can click and drag on the image to define a centroid region. Currently, the centroid region is rendered as anoverlay with a center marker, an aperture radius, and an inner and outer annulus ring that scale together using fixed ratios

In `ImageAnalysisView` we receive the centroid region updates and it stores the current region and tool state. Then it sends the centroiding request to the backend with the expected parameters. We either get the expected results or an error message.

## VISUALS
https://github.com/user-attachments/assets/28044cf9-1d0c-444b-a5c9-41c7372c8834

UPDATED SCREEN RECORDING

https://github.com/user-attachments/assets/fc8510a4-0131-4a45-85f0-9d021aa28617


src/components/Analysis/ImageViewer.vue
```diff
@@ -19,10 +19,27 @@ const props = defineProps({
     type: Object,
     required: false,
     default: null,
+  },
+  centroidRegion: {
+    type: Object,
+    required: false,
+    default: null,
+  },
+  centroidToolActive: {
```
jnation3406: why do we have both a prop and a separate `isCentroidToolActive` local variable? Seems like those could get out of sync. If we need bi-directional modification of the prop we can do that by setting it up with defineModel() and @update:blahModel

jnation3406 (approved):
Looks good, just one comment to simplify code using defineModel on the `centroidToolActive` prop for bi-directional updating. Also, the Centroid mode instructions on the right pane maybe should only show up when in centroiding mode, i.e. after the user clicks the tool button to enter that mode? Unless Rachel wanted it otherwise.

src/components/Analysis/ImageViewer.vue
capetillo: and you're so right

jnation3406 (approved):

PR #259, opened 2026-05-01
Operation: Light Curve

capetillo:
## Feature: Light Curve

Adds light curve as a single operation 

image

image

https://github.com/user-attachments/assets/d080220c-75f5-46d1-9cb4-bcb4f31667d3


jnation3406 (approved):
The code looks fine, but in your video, the operation wizard only shows a box for the target name, and not the ra/dec which should be populated when you click the search on the target name. I don't see how that PR affected this functionality, but if I look at current staging I still see it there so double check its not broken from this.

Image

PR #264, opened 2026-07-14
Feature/hr diagram

jnation3406:
This is the first draft of the HR diagram views. The structure of it should at least provide a strong starting point for implementing it if we don't want to use exactly what is here. 

Improvements to be made:

- There is a lot of special code for the interactivity on the plots (range selection on the distance and parallax plots, and ellipse selection and dragging on the proper motion plot). There is probably some room for improvement there to simplify things or rely on other libraries for that.
- The /scripts directory contains a python file that is used to query and generate the /public/parsec.json file used for the isochrone fitting. Since I committed that generated file, this script is just there if we every want to change the parameters and compute a new isochrone dataset.
-  It's currently showing a Parallax and Distance plot, but they are redundant so we should probably pick one and remove the other from the plot
- I think its too much info for a tablet. We probably should place the plots in tabs or something if the user is on a small/tablet screen so they can tab through the steps of picking pm/distance ranges, then fitting the isochrone on the CMD.
- I would like to add a button to set the initial isochrone distance param based on the distance ranges the user selects. This would get the initial distance much closer to what the user intended from their min/max selection range.


src/components/Analysis/DistanceHistogram.vue
```diff
@@ -0,0 +1,346 @@
...
+let distanceChart = null
+let themeColors = {}
+
+const BIN_COUNT = 30
+// keep a split-normal side from collapsing to a spike when a bound equals the estimate
+const MIN_LOG_SIGMA = 1e-3
+
+// Abramowitz & Stegun 7.1.26 error function (max error ~1.5e-7), for the split-normal CDF
+function erf(x) {
+  const t = 1 / (1 + 0.3275911 * Math.abs(x))
```
capetillo, line 52: can we declare these values as a const?

src/components/Analysis/DistanceHistogram.vue
```diff
@@ -0,0 +1,346 @@
...
+  const y = 1 - (((((1.061405429 * t - 1.453152027) * t) + 1.421413741) * t - 0.284496736) * t + 0.254829592) * t * Math.exp(-x * x)
```
capetillo, line 53: same here

src/components/Analysis/DistanceHistogram.vue
```diff
@@ -0,0 +1,346 @@
...
+// as an argument because the window plugin can fire during the Chart constructor's first draw,
+// before the component's chart variable is assigned
+function valueToPixel(chart, value) {
+  const { centers, logMin, logStep } = histogram.value
+  const area = chart.chartArea
+  const fraction = (Math.log10(value) - logMin) / (logStep * centers.length)
+  return area.left + fraction * (area.right - area.left)
+}
+
+function pixelToValue(chart, pixel) {
```
capetillo, line 138: Do you think this function could be used elsewhere? If so, it should be turned into a util function

src/components/Analysis/HRDiagramPlot.vue
```diff
@@ -0,0 +1,427 @@
...
+  hrDiagramChart.update()
+}
+
+function createChart() {
+  themeColors = getThemeColors()
+
+  const bounds = axisBounds()
+
+  const datasets = [buildStarDataset()]
+  if (props.isochrone && props.isochrone.length && props.isochroneVisible) {
```
capetillo, line 301: is this just an if statement? I think it should be contained

src/components/Analysis/IsochroneControls.vue
```diff
@@ -0,0 +1,192 @@
...
+  modelValue: {
+    type: Object,
+    required: true
+  },
+  // node values of the isochrone grid, sorted ascending
+  logAgeNodes: {
+    type: Array,
+    required: true
+  },
+  fehNodes: {
```
capetillo, line 24: what does feh mean?

capetillo (approved):
small comments regarding magic numbers. I think it looks good as a skeleton. There are a couple of parts here and there that are Claude-esque by having all the falllbacks. But it looks good!

src/components/Analysis/DistanceHistogram.vue
jnation3406: right now its specific to this because it uses a log10 scaled axis

src/components/Analysis/DistanceHistogram.vue
jnation3406: They dont really have descriptive names as they are literally coefficients of an error function defined in some paper

PR #268, opened 2026-07-20
Load the diagnostic images from s3 and show operation status message …

jnation3406:
…as tooltip

https://github.com/user-attachments/assets/a6611071-0829-4a07-aa48-bf5e821f2775

image


capetillo (approved):
lgtm!

PR #270, opened 2026-07-28
Allow clicking to expand and diagnostic images to full resolution and…

jnation3406:
… click again to shrink back. The attached video was pre resolution bump in the backend so it should be a little less pixelated then what this shows when full screen.

https://github.com/user-attachments/assets/8df4df2e-ccbb-4d54-b500-428b85f9e1a7


capetillo (approved):
nice!

PR #272, opened 2026-08-06
Initial set of improvements and refactor to the HR diagram views

jnation3406:
Here is a list of the things done in this PR:

- Refactor the HR Diagram specific stuff in DataAnalysisView to reside in HRDiagramOutput. This holds a tab window with 3 tabs for the 3 plots related to the HR Diagram output, storing the shared information between the plots within it. Removed usage of the Parallax plot but left the code there in case we want to re-add it later.
- Update selection and refactor for shared use in the distance/parallax/proper motion plots, and add touch screen friendly controls. This means larger grab radius on touch screens for edges, and allowing pinch/zoom/move with two fingers on the proper motion plot's circle selection.
- Optimize the updating of charts so they are much more performant for selecting cluster membership. Now it should be smooth even on tablets with thousands of stars.
- Update MultiImageInputSelector used when selecting input images for the operation wizard. Updated to take into account the max/min image requirements when not in color mode.
- If user hasn't manually changed isochrone distance modulus value, then if they update the distance histogram plot max/min range we will automatically update the isochrone distance modulus to the midpoint. This makes the initial isochrone match better with their intent on the distance of their cluster.


src/components/Analysis/DistanceHistogram.vue
```diff
@@ -170,6 +181,22 @@ const barColors = computed(() => {
   })
 })
 
+// the backend only suggests a distance window when the matched stars had usable distances
+const hasDefaults = computed(() => {
+  const guess = props.membershipGuess
+  return !!guess && Number.isFinite(guess.distance_min) && Number.isFinite(guess.distance_max)
```
capetillo: `Number.isFinite` smells like AI. Is there a scenario in which number would not be finite?

src/components/Analysis/DistanceHistogram.vue
```diff
@@ -300,6 +327,47 @@ onBeforeUnmount(() => {
...
         : 'Drag across the plot to select a distance window' }}
     </p>
+    <v-sheet
+      class="filter-controls pa-3 d-flex flex-wrap align-center ga-3"
+      color="var(--card-background)"
+      rounded
+    >
+      <v-number-input
+        :model-value="props.distanceMin"
+        label="distance min (pc)"
```
capetillo, line 337: what does pc stand for?

src/components/Analysis/HRDiagramOutput.vue
```diff
@@ -0,0 +1,378 @@
...
+  distance_min: null, distance_max: null
+}
+const membership = ref({ ...emptyMembership })
+
+/*
+  Every cut is independent and optional: filtering runs as soon as any one of them is set, and
+  a star has to pass each cut that is set.
+*/
+const pmCutActive = computed(() =>
+  ['pmra', 'pmdec', 'pm_radius'].every((field) => Number.isFinite(membership.value[field]))
```
capetillo: `Number.isFinite` isNecessary?

src/components/Analysis/HRDiagramOutput.vue
```diff
@@ -0,0 +1,378 @@
...
+)
+const distanceCutActive = computed(() =>
+  Number.isFinite(membership.value.distance_min) && Number.isFinite(membership.value.distance_max)
```
capetillo: same

src/components/Analysis/HRDiagramOutput.vue
```diff
@@ -0,0 +1,378 @@
...
+)
+// Parallax is currently not shown, but remains here if we want to add it later
+const parallaxCutActive = computed(() =>
+  Number.isFinite(membership.value.parallax_min) && Number.isFinite(membership.value.parallax_max)
```
capetillo: and same

capetillo: assume i always as this

src/components/Analysis/HRDiagramOutput.vue
```diff
@@ -0,0 +1,378 @@
...
+  we automatically update the distance modulus to be the midpoint of the range and update the isochrone.
+*/
+function setDistanceBound(field, value) {
+  membership.value[field] = value
+  const { distance_min: min, distance_max: max } = membership.value
+  if (muManuallySet.value || !isochroneFit.value) {
+    return
+  }
+  // a half-drawn or cleared window has no midpoint; leave the line where it is
+  const mu = Number.isFinite(min) && Number.isFinite(max) ? distanceModulusFromParsecs((min + max) / 2) : null
```
capetillo: mu?

src/utils/downloadChart.js
```diff
@@ -1,69 +1,69 @@
...
-      if (scale.ticks) scale.ticks.color = '#000'
-      if (scale.title) scale.title.color = '#000'
-      if (scale.grid) scale.grid.color = '#000'
-      if (scale.border) scale.border.color = '#000'
-    })
-  }
-
+  // Any lines after the first (the light curve sends source and aperture details) become the
+  // subtitle, small like the page shows them, so they don't eat into the plot's space
+  const [heading, ...subtitleLines] = Array.isArray(titleText) ? titleText : [titleText]
```
capetillo, line 22: Is there reason why `titleText` wouldn't be array?

src/utils/downloadChart.js
```diff
@@ -1,69 +1,69 @@
...
   chart.update('none')
 
   // Wait for the chart to finish rendering
   await new Promise(resolve => setTimeout(resolve, 100))
 
-  // Export as PNG
-  let pngUrl = chart.toBase64Image('image/png', 1)
+  // Export as PNG: page background, then the chart canvas, then any hand-drawn legend
+  const scale = chart.currentDevicePixelRatio || 1
+  const legendWidth = legendItems.length ? 190 * scale : 0
```
capetillo, line 46: magic number 190. but why 190?

src/utils/downloadChart.js
```diff
@@ -1,69 +1,69 @@
...
-    legendCanvas.height = chart.canvas.height
-    ctx.drawImage(chart.canvas, 0, 0)
     ctx.font = `${16 * scale}px sans-serif`
     ctx.textBaseline = 'middle'
+    ctx.strokeStyle = textColor
 
     legendItems.forEach((item, index) => {
       const y = legendY + index * 26 * scale
       ctx.fillStyle = item.color
       ctx.fillRect(legendX, y - 6 * scale, 12 * scale, 12 * scale)
```
capetillo, line 66: couple of magic numbers here

src/utils/downloadChart.js
```diff
@@ -1,69 +1,69 @@
...
-      ctx.fillStyle = '#000'
+      ctx.fillStyle = textColor
       ctx.fillText(item.label, legendX + 20 * scale, y)
```
capetillo, line 68: and here

capetillo (approved):
looks good, I would just search for `Number.isFinite` and see if that's necessary as well as `Array.isArray`. A couple of magic numbers here and there, but it all looks good!

src/components/Analysis/DistanceHistogram.vue
jnation3406: parsec, its the units of distance

src/components/Analysis/HRDiagramOutput.vue
jnation3406: the greek letter. Its the notation used in the HR diagram isochrone matching thing to denote distance modulus

src/utils/downloadChart.js
jnation3406: Most of the times we use it its just a string for the title. There is only one case in the LightCurvePlot where it passes an array since it has a subtitle and a title - the rest just have a title only and pass a string.

src/components/Analysis/HRDiagramOutput.vue
capetillo: ah!

src/utils/downloadChart.js
jnation3406: I guess I'm kind of fine with these. They are pixel values for drawing parts of the legend and are just used in one line so I don't think its worth adding constants for each part of each thing we draw in pixels.

